Issue #87, opened 2019-07-15
Increase the PWM frequency in order to reduce the current ripple due to PWM modulation


nunoguedelha:
We suggest here an improvement for reducing the motor current ripples due to the PWM duty-cycle modulation.
The PWM ON/OFF phases result in a current oscillation which can reach 100mA amplitude, as initially analyzed and described in [this issue](https://github.com/dic-iit/element_torque-control-via-current/issues/61):

> The 100mA amplitude noise is actually the normal oscillation due to the PWM voltage excitation filtered by the phase inductance (~0.200mH). The same results can be obtained in a simulation where we feed with a PWM a R-L circuit having the same R, L values specified in the motor MOOG C2900584 datasheet as an example. Actually, as per prior tests run by @maggia80 's team, a single phase would exhibit an oscillation of ~1A peak-to-peak (I have a copy of the tests but not a link). But in a 3 phases Star configuration, the combination of the 3 respective PWM sources results in shorter PWM "ON" cycles, and so, smaller current ripples.

- The current ripples can be reduced by increasing the PWM frequency (above 20 kHz), regardless of the frequency of the Field Oriented Control task running on the FOC (20 kHz)
- This should not affect the settling time of the current as shown in the figure below (refer to section 2.3.3 and Fig.2.16 of my PhD dissertation):

### PWM freq = 20 kHz
![pwm2currentRipple-20khz-pwm20perc](https://user-images.githubusercontent.com/6848872/61222676-73301f80-a71b-11e9-9b9d-297fbc4d3577.png)
Current ripple amplitude = 2 A

### PWM freq = 80 kHz
![pwm2currentRipple-80khz-pwm20perc](https://user-images.githubusercontent.com/6848872/61222707-88a54980-a71b-11e9-9779-37e9be3f752a.png)
Current ripple amplitude = 0.5 A

nunoguedelha:
In order to proceed with the implementation we need to verify the following points:
- [ ] have we tried before to increase the PWM and/or the DMA interrupt task frequency to 40kHz?
- [ ] what is the maximum frequency available?
- [ ] where is this properly documented (datasheet, section,...)?
- [ ] I could find the datasheet of the dsPIC33FJ128MCX02 in the Microchip site [here](http://ww1.microchip.com/downloads/en/devicedoc/70291g.pdf). That datasheet does not explain clearly how to configure the PWM frequency. Do we have other documentation on this?

[...]

PR #90, opened 2019-08-07
Increased PWM freq to 40kHz

nunoguedelha:
Implements https://github.com/robotology/icub-firmware/issues/87. List of changes:
- set PWMFREQUENCY define.
- re-scale Vq, Vd with respect to the PWM resolution and fullscale.
- fixed PTPER computation.
- set P1SECMPbits instead of SEVTCMPbits.


nunoguedelha:
The following **non-regression tests** have been run on iCubGenova04 (Green iCub):
- balancing on two feet with left-right trajectory of the center of mass.
- Yoga on left foot (balancing on one foot and moving right leg and arms).
These tests were run while the controllers were configured in PWM output mode, the 2FOC running at 40kHz.
The Current Control mode (on yarpmotorgui) has been tested on one of the legs (left).

All tests passed.

PR #91, opened 2019-08-14
Added 2FOC V3.5 application.

ale-git:
Now the 2.5 motors are controlled in a way as much similar as possible to 3.0 motors in the experimental release V3.5 of the 2FOC firmware:

- the quadrature encoder alone is used to measure the rotor angle
- the Hall effect sensors are used to reset the quadrature encoder (1 <-> 6 sector transition), in the same way the Index tick is used with 3.0 motors.


ale-git:
Some improvements have been added:

- now the integer arithmetic of the Clarke-Park transformations is more precise thanks to rounding terms
- gain calibration of the ADC converter is now performed at startup in addition to offset calibration

nunoguedelha:
We have to test on the robot before merging the PR.

nunoguedelha:
@ale-git why have we duplicated the whole folder? Are maintaining 2 releases? If it's the case it's better to create release branches. It makes it easier to compare/merge changes between the 2 releases.

nunoguedelha:
## Moving the files from 2FOC-V35 to 2FOC-V3
@ale-git I ported the changes into the original folder `2FOC-V3`, merging them with the changes for:
- the PWM clock rate increase
- the fix of the memory alignment issue.

Once the review is finished, we can delete the files in `2FOC-V35` folder.

@marcoaccame, as agreed with @ale-git , we moved the files instead of keeping duplicates, for the following reasons:
- keeping a clear history of all the changes so far
- if eventually we keep two releases V3.5.x and V3.3.x (I strongly suggest not to), it will be easier to deploy common fixes to both releases and track those changes. For that we need to use two release branches, but that can be discussed in a next meeting..

I suggest we discuss on this today on skype (I'm working from home today and tomorrow morning), or tomorrow afternoon.

nunoguedelha:
We agreed to revert the HES/ADC offset calibration to the former original method: PWM on all phases is set to OFF prior measuring the offset currents on terminals A and C. Refer to https://github.com/robotology/icub-firmware/issues/96#issue-487603910.

Changes pushed in 41b228b. But this change was done on the old folder 2FOC-V35. Porting now to 2FOC-V3.

nunoguedelha:
ciao @ale-git , here are the tests I think would be performance improvement indicators for this pull request, illustrating the two main expected improvements:
1. obtain a clean current waveform on each stator phase, without current discontinuities through a full electric cycle.
2. have a stator direct field (Id) always well aligned with the rotor, even after a few laps and transitions between sectors 1 & 6.

**Here's a proposal for the tests to run...**

## Test 1: clean current waveform on each stator phase
This test is performed using the oscilloscope (I can help you on this). We have to verify that there are no current discontinuities through a full electric cycle. These discontinuities were previously observed at the switching between sectors 1<->2<->3<->4<->5<->6, while the rotor position detection was hybrid, using the optical encoder and the HES.
With the new solution you've implemented, there could still be a small step, but just between sectors 1 & 6, when the variable tracking the rotor position is reset.

## Test 2: stator direct field (Id) always well aligned with the rotor
After a few laps and switching between sectors 1 & 6, there could be a slight misalignment between the stator field and the rotor. We can measure that error through the following procedure (while the rotor has no mechanical load):
1. run the usual rotor alignment procedure,
2. set Iq = 0, and wait a few ms that the rotor position stabilizes (cogging effects),
3. stop the rotor position tracking (`delta` is fixed),
4. set Id to a high value. If there is a misalignment, the rotor moves a fraction or a few degrees.
5. measure final rotor position and compare to `delta`.

@ale-git, does this seem reasonable with respect to the new rotor alignment you have implemented?

marcoaccame:
having talked w/ @ale-git we can close this old PR w/out merging

PR #113, opened 2020-04-01
strain2-FT-streaming-at-1khz

marcoaccame:
This PR enables the strain2 to stream FT values at 1KHz.

That is achieved by setting the CPU clock at faster speed (`80MHz` instead of 16MHz) and using a slightly large averaging time for ADC (`ADC_SAMPLETIME_12CYCLES_5` instead of ADC_SAMPLETIME_6CYCLES_5)  which reduce the averaging time of the ADC peripheral of a factor 2.5 which:
- allows to complete acquisition of the six ADC values of the strain gauges to about 550 usec (it was abouth 1100 usec) allowing in such a way the tx rate to be every 1000 usec.
- still allows a good averaging for noise removal

Comprehensive tests have been done on:
- the correctness of the acquired values even with a reduced averaging time (brief of which you can read below)
- coexistence of the FT @ 1 KHz with other sources of streaming (e.g., w/ IMU).

image

image

image


traversaro:
fyi @fjandrad @isorrentino @nunoguedelha @GianlucaMilani

PR #117, opened 2020-07-24
feat/timesynch

marcoaccame:
This PR is associated with a [PR](https://github.com/robotology/icub-firmware-shared/pull/38) in `icub-firmware-shared` with the same name. 

The purpose of these two joint PRs is to give to the ETH boards the possibility to synchronize their clock reference by means of suitable messages which carry the desired time in micro-seconds format. 

See the [PR](https://github.com/robotology/icub-firmware-shared/pull/38) for more details about what code is in `icub-firmware-shared`.

In `icub-firmware` there is the use of such messages by the boards `ems`, `mc4plus`, `mc2plus`.

The boards, by default, are able to decode them when they arrive to port 12345, which is the one used by `yarprobotinterface`. The produced binary has been tested to work successfully with  `yarprobotinterface` even if , so far, it does not manage the sending of these messages. 

However, by special macro definition, it is possible to produce a binary which is able to decode the messages also when they arrive on port 11000. This port can be used by any process, maybe dedicated to the synchronization task of the embedded boards.

One of such processes is the `diagnostic-daemon`, which can be always running and is able to broadcast in the ETH network to port 11000 a time synchronization beacon which uses the `set<..microsecondvalue>` message with time aligned to Linux/YARP/NTP server. Details on the `diagnostic-daemon` will soon be available.

Just after the merging of this PR and of the [PR](https://github.com/robotology/icub-firmware-shared/pull/38), I shall produce the relevant binaries `ems.hex`, `mc4plus.hex`, `mc2plus.hex` in `icub-firmware-build:devel` able to decode from port 12345 and also from port 11000.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/ems004/appl/v2/cfg/eoemsappl/EOMtheEMSapplCfg_cfg.h
+++ b/emBODY/eBcode/arch-arm/board/ems004/appl/v2/cfg/eoemsappl/EOMtheEMSapplCfg_cfg.h
@@ -74,7 +74,7 @@ extern "C" {
 //  <o> major           <0-255> 
 #define EOMTHEEMSAPPLCFG_VERSION_MAJOR          3
 //  <o> minor           <0-255> 
-#define EOMTHEEMSAPPLCFG_VERSION_MINOR          31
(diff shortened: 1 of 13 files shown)


traversaro:
Fyi @robotology/iit-dynamic-interaction-control

PR #120, opened 2020-08-06
feat/adcstrain2-get-all

marcoaccame:
**Basic Description**

This PR introduces a backward-compatible enhancement for the CAN message `GET_CH_ADC` when managed by board `strain2`.

If the field CHN of this message is 0x0f, then the board `strain2` interprets it as `StrainChannel::all` and sends back a burst of six messages with the values of the size ADC channels. 

For more details see section `3.3.2 3.3.2 Messages for managing the FT and HES data service` of document ....

**Use case**

An agent wants to retrieve in polling mode the 6 values of the ADC with minimal jitters time.

In the past, one had to send six separate `GET_CH_ADC ` commands. That resulted in a complete operation executed in longer time (> 3 ms) and with values coming from separated ADC acquisitions. If one wanted values from a single acquisition teh only ways was to use the streaming mode with messages such as `set tx rate`, `start tx` and eventually `stop` message.

Now a single request is enough. All values come from a single ADC acquisition and the measured jitters time as measured w/ CANreal is 920 usec. This time is equal to time to send the single `GET_CH_ADC ` request +  ADC acquisition time + time to send back the six replies.

See picture where the request command is highlighted.

image

**Compatibility issues**

The request message `GET_CH_ADC ` keeps the same format. It only has an extra possible value in field CHN. If this values is 0xF it is normally discarded by every board but `strain2` boards w/ version >= 2.08 (soon in `icub-firmware-build:devel`). Values from 0 to 5 are normally managed as before.

The reply messages are unchanged.


marcoaccame:
added the [binary for the `strain2` board](https://github.com/robotology/icub-firmware-build/commit/f11975c9276c855fb5e18ef004c63ae22ecbe5fa)

PR #132, opened 2020-09-23
MOM SECURITY.md

KCknows:
Why has this happened? Why is Embot not responding? I want ALL the bugs identified and pulled out now! 
This is A REAL LIFE! Do you people understand how important all of this is!!! Embot is a little girl in real life!!! Omg :(


KCknows:
Where is she?

PR #153, opened 2020-12-15
2foc firmware - fix compilation error. 

### LINKED ISSUE #152: Fix building problem - MPLAB
When building the 2FOC.V3 project, the following error appears:

```../src/ecan.c:827:37: error: conflicting types for 'int8_t'```

 It is a problem related to the new compiler version. The old version contains this line:

https://github.com/robotology/icub-firmware/blob/b6c0a716bb75105fd1d62cfce6ae05c7d649df9b/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c#L827

that causes that error.

The fix is:

https://github.com/isorrentino/icub-firmware/blob/afd340abb343c5d5502970961266feaac327dd8a/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c#L1135

isorrentino:
This fix #152.

cc @giulioromualdi @traversaro @nunoguedelha @marcoaccame

### DIFF
--- a/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c
+++ b/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c
@@ -824,7 +824,7 @@ void __attribute__((interrupt, no_auto_psv)) _DMA2Interrupt(void)
 
 //VALE
 
-typedef     char                    int8_t;
+typedef     signed char             int8_t;
 typedef     unsigned char           uint8_t;
 typedef     signed int              int16_t;
 typedef     unsigned int            uint16_t;


marcoaccame:
Hi @ale-git, can you pls have a look at that?

marcoaccame (changes requested):
pls address this issue:
- i have spotted that the PR is vs master. we typically do development in devel, hence most PR are vs devel unless there is some nasty bug which we want to solve and which affects robots around. is there a reason for you having used master? if not, would it be possible to use devel instead?

about the content of the PR: 
- i would rather remove all the 8 lines w/ the `typedef` and put a nice and standard `#include "stdint.h"` instead. can you pls test if it solves the compilation problem?

isorrentino:
Hi @marcoaccame, sorry for the late reply.

> i have spotted that the PR is vs master. we typically do development in devel, hence most PR are vs devel unless there is some nasty bug which we want to solve and which affects robots around. is there a reason for you having used master? if not, would it be possible to use devel instead?

There is no reason in using master instead of devel, so I will rebase my branches on devel. Thank you for the information.

> i would rather remove all the 8 lines w/ the typedef and put a nice and standard #include "stdint.h" instead. can you pls test if it solves the compilation problem?

I will test the change as soon as possible.

isorrentino:
cc @giovannipizzolante

giovannipizzolante:
I have open another PR vs devel where I removed all the 8 lines w/ the ```typedef``` and put a nice and standard ```#include "stdint.h"``` instead and this solves compilation problem.

traversaro:
@isorrentino @giovannipizzolante I think we can close this PR as it has been superseded by https://github.com/robotology/icub-firmware/pull/158, right?

isorrentino:
Yes. Closing.

PR #158, opened 2021-01-15
 2FOC Firmware -  fix compilation error

### LINKED ISSUE #152: Fix building problem - MPLAB

giovannipizzolante:
This fix #152 . As suggested by @marcoaccame the compilation error has been fixed by removing ```typedef``` lines and putting an ```#include "stdint.h"``` .

### DIFF
--- a/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c
+++ b/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c
@@ -6,7 +6,7 @@
 
 #include <p33FJ128MC802.h>
 #include <string.h>
-
+#include "stdint.h"
 #include "ecan.h"
 #include "system.h"
 #include "faults.h"
@@ -823,15 +823,6 @@ void __attribute__((interrupt, no_auto_psv)) _DMA2Interrupt(void)
 
 
 //VALE
-
-typedef     char                    int8_t;
-typedef     unsigned char           uint8_t;
-typedef     signed int              int16_t;
-typedef     unsigned int            uint16_t;
-typedef     signed long             int32_t;
-typedef     unsigned long           uint32_t;    
-typedef     signed long long        int64_t;
-typedef     unsigned long long      uint64_t; 
 
     
 extern void hal_can_receptionfilter_set(unsigned char mask_num, unsigned long mask_val, unsigned char identifier_num,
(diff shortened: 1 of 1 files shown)


emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c
```diff
@@ -6,7 +6,7 @@
 
 #include <p33FJ128MC802.h>
 #include <string.h>
-
+#include "stdint.h"
```
traversaro, line 9: Thanks @giovannipizzolante , I have a nitpick (i.e. a comment that you can safely ignore, see https://google.github.io/eng-practices/review/reviewer/looking-for.html#style). 

In C/C++, the main difference between includes that are added with `<include-name.h>` vs the one that are included with  `"include-name.h"` is that the latter one, i.e. the one with the quote, also looks in the same directory of this file to search for the included .h file (see https://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename). For this reason, the style with which normally system includes are added is the one with the `<` , `>`. 
```suggestion
#include <stdint.h>
```

traversaro (approved):

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/ecan.c
giovannipizzolante: thanks @traversaro !

traversaro:
@marcoaccame Feel free to let @giovannipizzolante if you prefer brackets or commas for the include (see https://github.com/robotology/icub-firmware/pull/158#discussion_r558293317), thanks!

marcoaccame:
Hi @traversaro: thanks for the suggestion. i have committed it. an thanks @giovannipizzolante for the test.

PR #170, opened 2021-02-19
Implement error counter so the logs don't overflow

AlexAntn:
This PR introduces an error counter and corresponding threshold to prevent an overflow of the logger when there is a non-HW-fault related issue.

Current behavior sends an error message up to 10 times, then stops sending the same error. 

This counter is reset if/when the error disappears (allowing to see if the error is consistent during an experiment, or just pops up a couple of times).

Please review code.

CC @pattacini @S-Dafarra

### DIFF
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
@@ -269,6 +269,7 @@ void Motor_init(Motor* o) //
     o->fault_state_prec.bitmask = 0;
     o->fault_state.bitmask = 0;
     o->diagnostics_refresh = 0;
+    o->err_counter = 0;
     
     WatchDog_init(&o->control_mode_req_wdog);
     WatchDog_init(&o->can_2FOC_alive_wdog);
@@ -607,6 +608,7 @@ void Motor_force_idle(Motor* o) //
     o->fault_state_prec.bitmask = 0;
     o->fault_state.bitmask = 0;
     o->diagnostics_refresh = 0;
+    o->err_counter = 0;
     
     o->hardware_fault = FALSE;
 }
@@ -622,6 +624,7 @@ static void Motor_send_error(uint8_t id, eOerror_value_MC_t err_id, uint64_t mas
     eo_errman_Error(eo_errman_GetHandle(), eo_errortype_error, NULL, NULL, &descriptor);    
 }
 
+#define ERROR_COUNTER_MAX                 10
 
 BOOL Motor_check_faults(Motor* o) //
 {
@@ -673,12 +676,24 @@ BOOL Motor_check_faults(Motor* o) //
         }
 
         // No HW fault triggered by this warnings
-        if (++o->diagnostics_refresh_warning > 5*CTRL_LOOP_FREQUENCY_INT) {
-            if (o->qe_state.bits.dirty) {
-                Motor_send_error(o->ID, eoerror_value_MC_motor_qencoder_dirty, 0);
+        if (++o->diagnostics_refresh_warning > 5*CTRL_LOOP_FREQUENCY_INT ) {
+            if (o->err_counter < ERROR_COUNTER_MAX) // we are still below the maximum number of error messages
+            {
+                if (o->qe_state.bits.dirty) {
+                    Motor_send_error(o->ID, eoerror_value_MC_motor_qencoder_dirty, 0);
+                    ++o->err_counter;
+                }
+                if (o->qe_state.bits.phase_broken) {
+                    Motor_send_error(o->ID, eoerror_value_MC_motor_qencoder_phase, 0);
+                    ++o->err_counter;
+                }
             }
-            if (o->qe_state.bits.phase_broken) {
-                Motor_send_error(o->ID, eoerror_value_MC_motor_qencoder_phase, 0);
+            else // if the maximum number of error messages has been received
+            {
+                if (!o->qe_state.bits.dirty && !o->qe_state.bits.phase_broken) // if not more errors detected, reset counter
+                {
+                    o->err_counter = 0;
+                }
             }
         }
     }
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
@@ -269,7 +269,8 @@ void Motor_init(Motor* o) //
     o->fault_state_prec.bitmask = 0;
     o->fault_state.bitmask = 0;
     o->diagnostics_refresh = 0;
-    o->err_counter = 0;
+    o->diagnostics_refresh_warning = 0;
+    o->qencoder_err_counter = 0;
     
     WatchDog_init(&o->control_mode_req_wdog);
     WatchDog_init(&o->can_2FOC_alive_wdog);
@@ -608,7 +609,8 @@ void Motor_force_idle(Motor* o) //
     o->fault_state_prec.bitmask = 0;
     o->fault_state.bitmask = 0;
     o->diagnostics_refresh = 0;
-    o->err_counter = 0;
+    o->diagnostics_refresh_warning = 0;
+    o->qencoder_err_counter = 0;
     
     o->hardware_fault = FALSE;
 }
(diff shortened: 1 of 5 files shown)


emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.h
```diff
@@ -218,6 +218,8 @@ typedef struct //Motor
 
     HardStopCalibData hardstop_calibdata;
 
+    uint16_t err_counter;
```
pattacini, line 221: I would give the variable a more expressive name; something like `qencoder_err_counter`

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
```diff
@@ -673,12 +676,24 @@ BOOL Motor_check_faults(Motor* o) //
...
+                    ++o->err_counter;
+                }
             }
-            if (o->qe_state.bits.phase_broken) {
-                Motor_send_error(o->ID, eoerror_value_MC_motor_qencoder_phase, 0);
+            else // if the maximum number of error messages has been received
+            {
+                if (!o->qe_state.bits.dirty && !o->qe_state.bits.phase_broken) // if not more errors detected, reset counter
+                {
+                    o->err_counter = 0;
```
pattacini, line 695: Ideally, we would need to communicate to the higher level that the error condition has disappeared (to make an info message show up in `yarplogger`).

Is there such a possibility? Do we have already the required infrastructure for sending out info messages?
To be agreed whether this can be done easily.

cc @marcoaccame

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
```diff
@@ -673,12 +676,24 @@ BOOL Motor_check_faults(Motor* o) //
         }
 
         // No HW fault triggered by this warnings
-        if (++o->diagnostics_refresh_warning > 5*CTRL_LOOP_FREQUENCY_INT) {
-            if (o->qe_state.bits.dirty) {
-                Motor_send_error(o->ID, eoerror_value_MC_motor_qencoder_dirty, 0);
+        if (++o->diagnostics_refresh_warning > 5*CTRL_LOOP_FREQUENCY_INT ) {
```
pattacini, line 679: `diagnostics_refresh_warning` was not initialized correctly and this seems to be still the case.

pattacini (changes requested):
Thanks @AlexAntn 👍🏻 

I have got a few comments on this.

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
AlexAntn: I have updated the code to both initialize it properly, and reset the value when the warning is triggered

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.h
AlexAntn: updated the variable name

pattacini (approved):

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
pattacini: I would keep this message still in the conversation in order to collect ideas and suggestions/recommendations.

marcoaccame: Hi, we surely can send an info message which ends up in `yarplogger`. 

here is an example which sends a message which informs that the error `eoerror_value_MC_motor_qencoder_phase` has disappeared. This code will print the string `MC: qenc error has disappeared`, but you can have what you want.

```C++
#include "EoError.h"

static void Motor_send_diagnostic(eOerrmanErrorType_t errortype, uint8_t id, eOerror_value_MC_t err_id, uint64_t mask);


void somefunction(void)
{
    ... 

    Motor_send_diagnostic(eo_errortype_warning, o->ID, eoerror_value_MC_motor_qencoder_phase_disappeared, 0);

}

static void Motor_send_diagnostic(eOerrmanErrorType_t errortype, uint8_t id, eOerror_value_MC_t err_id, uint64_t mask)
{
    static eOerrmanDescriptor_t descriptor = {0}; 
    // but also not static can be fine as well. as long as we have enough stack size. 
    // eOerrmanDescriptor_t is 16 bytes big    
    descriptor.par16 = id;
    descriptor.par64 = mask;
    descriptor.sourcedevice = eo_errman_sourcedevice_localboard;
    descriptor.sourceaddress = 0;
    descriptor.code = eoerror_code_get(eoerror_category_MotionControl, err_id);
    eo_errman_Error(eo_errman_GetHandle(), errortype, NULL, NULL, &descriptor);    
}
```

But you also need to work on files `EoError.h` and `EoError.c` in `icub-firmware-shared`.

In particular, you must add a new item in `eOerror_value_MC_t`

```C++
    eoerror_value_MC_motor_qencoder_phase_disappeared = 17
} eOerror_value_MC_t;

enum { eoerror_value_MC_numberof = 18 };

``` 

and create a new associated string in the lookup table:

```C++
    {eoerror_value_MC_joint_hard_limit,      "MC: hard limit reached. The joint position is outside its hardware boundaries. par16 = ID of joint."},
    {eoerror_value_MC_motor_qencoder_phase_disappeared,      "MC: qenc error has disappeared"}
 
};  EO_VERIFYsizeof(eoerror_valuestrings_MC, eoerror_value_MC_numberof*sizeof(const eoerror_valuestring_t)) 

```

pattacini: Fantastic! Thanks @marcoaccame for the how-to guide!

pattacini (changes requested):
@AlexAntn I would add up the info message following @marcoaccame 's how-to guide.
Needs to be carefully tested any how.

marcoaccame:
> @AlexAntn I would add up the info message following @marcoaccame 's how-to guide.
Needs to be carefully tested any how.

It is also required a small change in icub-firmware-shared, possibly w/ increase of its [version](https://github.com/robotology/icub-firmware-shared/blob/devel/CMakeLists.txt#L9) and another one in icub-main so that it is forced to use the new version of icub-firmware-shared: [here](https://github.com/robotology/icub-main/blob/devel/conf/iCubFindDependencies.cmake#L37).

pattacini:
>It is also required a small change in icub-firmware-shared, possibly w/ increase of its version and another one in icub-main so that it is forced to use the new version of icub-firmware-shared: here.

Yep, indeed we agreed on waiting until distro 2021.02 is out and the rush is over given the complexity of such a change.

AlexAntn:
I have updated this PR and opened this PR on icub-firmware-shared with the changes for the info message: [https://github.com/robotology/icub-firmware-shared/pull/46](https://github.com/robotology/icub-firmware-shared/pull/46).

I have also implemented the changes on icub-main (to force the last version, for testing), but I have only committed them to my fork of icub-main, should I also open a PR?

pattacini:
>I have also implemented the changes on icub-main (to force the last version, for testing), but I have only committed them to my fork of icub-main, should I also open a PR?

Sure! Thanks @AlexAntn 
The PR shall target `icub-main@devel`.

pattacini (approved):

pattacini:
Here's the PR on icub-main: https://github.com/robotology/icub-main/pull/723.

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
pattacini: @marcoaccame could you please review what @AlexAntn did following your hints?

marcoaccame (commented):
Hi @AlexAntn, you wrote:
> We are testing these changes thoroughly to make sure nothing is affected.

Did you manage performing your test? If so, let me know. The changes seems correct to me. If tests are OK we can merge.

AlexAntn:
I have managed to test the firmware with the single motor setup we have, and the mechanism for sending the messages is working as intended (send 10 messages, stop, reset when the error is gone and inform the user).

The only issue I am currently trying to understand has to do with the actual string sent in the Info message that should inform the user that the error has disappeared. Regarding this issue, I have realized the following:

- I implemented a new `eoerror_value_MC_motor_qencoder_phase_disappeared` value for the info message, which has the value 17 (incremented the `enum` to account for this new value) in [these lines](https://github.com/robotology/icub-firmware-shared/blob/964d2736ab515ad4083c5f8e832c5933add3ae46/eth/embobj/plus/comm-v2/icub/EoError.h#L223-L226)
- I implemented also the corresponding string [here](https://github.com/robotology/icub-firmware-shared/blob/964d2736ab515ad4083c5f8e832c5933add3ae46/eth/embobj/plus/comm-v2/icub/EoError.c#L195-L196)
- When I try sending this code through the `Motor_send_diagnostic` function (defined [here](https://github.com/robotology/icub-firmware/blob/55c8f07cfe3ca46164a49bc38f3de03ce1e4ee44/emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c#L629-L640) and called [here](https://github.com/robotology/icub-firmware/blob/55c8f07cfe3ca46164a49bc38f3de03ce1e4ee44/emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c#L711)) it prints the following message: 
```
from BOARD 10.0.1.1 (░[m:), src LOCAL, adr 0, time 2057s 743m 838u: (code 0x02000011, par16 0x0000 par64 0x0000000000000000) -> eoerror_INTERNALERROR__stringatwrongplace + .
```
After looking at the code, I figured this is because the actual value being used to query for the message string is higher than the maximum number of entries (18). 
- I tried re-ordering the entries on both the enum of values and the list of messages, and found that when I use the value `eoerror_value_MC_motor_qencoder_phase_disappeared` it always accesses the message string corresponding to the subsequent entry (basically, like it was incremented by 1). This would explain the value being higher than the maximum (since this entry was the last one in the enum);
- Furthermore, I noticed that when I place this entry somewhere in the middle of the list (and re-order all entries accordingly), this error propagates to all the entries that come after this one (in that when I use them to send a message, it takes the value of the subsequent entry like it was incremented by 1).

I have tried getting to the bottom of this, but debugging at this level is proving to be challenging. Any possible hints on what could be generating this error, @marcoaccame (or others)?

pattacini:
Thanks a lot @AlexAntn for the very accurate test 👍🏻
So, we have potentially hit a lurking bug that impacts the handling of the info messages.

AlexAntn:
After a quick chat with @marcoaccame , I double-checked the configuration of icub-main, and tested again, and the message is now correctly displayed!

This completes the pending tests, the PR can go forward!

pattacini:
That's great!
Let's upload the build as the final step, while I'll be merging all the involved PR's.

PR #173, opened 2021-02-23
Added cable management to hand Mk3, activated by FINGER_MK3 macro

ale-git:
This PR is required to avoid too much stress on the steel tendons which drive the open-close movements of the hand mk3.

Its code is enabled only if macro `FINGER_MK3` is defined. This macro is defined only in a special project, `mc4plus.handv3.uvprojx` used only for the new hand.

If not defined, the code is exactly the same as the one running on actual iCub's boards.


marcoaccame (approved):
Hi @ale-git: the PR is fine, thanks.
Hi @Nicogene: this PR does not need any new build for the ETH boards and does not have impact on the imminent iCub release

Issue #174, opened 2021-03-02
Analysis of FW of mc4plus to solve board's disappearance from the ETH network


marcoaccame:
Following issue in [here](https://github.com/robotology/icub-tech-support/issues/673), we have identified a critical behaviour that should be investigated and solved:  some `mc4plus` boards disappears from the ETH network when running `yarprobotinterface`.

What happens is that an `mc4plus` board which offers the motion control service may stop transmitting and receiving UDP frames. The board maintains PWM actuation and seems to be stuck: it does not respond to `ping` and is not seen by `FirmwareUpdater`.

We have observed the board `10.0.1.24` to have such a misbehavior but also the board `10.0.1.20`. 

A careful analysis is required using multiple investigation paths. Amongst them:
- analysis of code differences in the motion control between `mc4plus` and `ems` in an attempt to spot bugs in the differences;
- attempt to recreate the misbehavior on a simplified setup, with the debugger connected to the `mc4plus` board so that we can trace exact behaviour;
- investigate if the misbehavior is possible also on a similar setup with the `ems` board;
- investigate if the misbehavior is possible also w/ other services such as F/T streaming etc;
- validate the behaviour / solutions found / tested on the simplified setup also on the robot.

cc: @S-Dafarra @DanielePucci @maggia80

marcoaccame:
An idea for the simplified setup is:
- a linux pc which communicates via ETH to an mc4plus / ems board by sending UDP frames which simulates the **repeated** activation / deactivation of the motion control service (and also others) in the same way as `yarprobotinterface` does.
- an mc4plus / ems board with attached a ULINK Pro debugger.
- presence of sensors and actuators possibly simulated.

marcoaccame:
Here are some news: **I can now systematically reproduce the bug on a dedicated setup**.

I have prepared the simplified setup which is able to run  4 possible tests: either the `FT service` or the `MC service` on either the `ems` or the `mc4plus`. 

For  the `FT service` I have used a true FT sensor on CAN1 (the `strain2`), whereas for the `MC service` I have just emulated sensors and did not use any actuator. I emulated the AEA sensors, I did not put any 2foc boards on CAN (but I added the `strain2` to avoid the overflow of the FIFO of CAN frames) nor I attached any DC motor.

[...]

PR #176, opened 2021-03-22
Multiple start of MC service

marcoaccame:
As investigated in https://github.com/robotology/icub-firmware/issues/174, the ems, mc4plus and mc2plus need a fix which solves their disappearance when yarprobotinterface is started / stopped several times.

This PR solves the problem.

As described in https://github.com/robotology/icub-firmware/issues/174#issuecomment-804266518, the change was extensively tested, so it is ready to be merged.

A dual [PR](https://github.com/robotology/icub-firmware-build/pull/25) on icub-firmware-build is available so that we can merge both of them at the same time.


marcoaccame:
tested the binaries once more (see https://github.com/robotology/icub-firmware-build/pull/25#issuecomment-806582432)

PR #180, opened 2021-04-13
Add MBD-based FT FW 

triccyx:
Used MBD generated code from Simulink as new FT sensor code.
Check Simulink model in https://github.com/robotology/icub-firmware-models/pull/2.


pattacini (commented):
I took the chance to trace back the pristine memory areas from which we do the `memcpy`:

https://github.com/robotology/icub-firmware/blob/b6c0a716bb75105fd1d62cfce6ae05c7d649df9b/emBODY/eBcode/arch-arm/embot/app/embot_app_application_theSTRAIN.cpp#L270-L271

Then, it'd be sufficient to move the `memcpy` into those code lines where `transfQ15matrix` and `transfQ15tare` are filled in to obtain the correct asynchronous update.

However, I see (and I remember now) that the FW deals with a set of diverse matrices and offsets; therefore, it might be convenient to keep things as they are now, for the sake of simplicity.

After all, the continuous copy does not take long.

triccyx:
> ... sake of simplicity

Agree

pattacini (approved):
@simeonedussoni and @triccyx tested the MBD FW today successfully 🚀 

Hence, @marcoaccame you can now review the code.

marcoaccame:
> Hence, @marcoaccame you can now review the code.

Hi @pattacini and @triccyx, I'm having a look at it.

marcoaccame (approved):
Well done: the changes are OK and can be merged. 



There are only some minor comments reported in the following.

- Inside `bool embot::app::application::theSTRAIN::Impl::processing()`  you moved the calibration inside   `if(false == runtimedata.data.adcsaturation) { .. }` . It was originally outside. Does this movement of code change the behaviour? I have checked. The behaviour is the same. I am telling that only because at first I thought there was a mistake. But it there is not any.



- When you produce a new binary, you will need to advance the application number so that we can identify the new binary vs current released version which now is:

  ```C++
  constexpr embot::app::theCANboardInfo::applicationInfo applInfo 
  { 
      embot::prot::can::versionOfAPPLICATION {2, 0, 10},
      embot::prot::can::versionOfCANPROTOCOL {2, 0}    
  };
  ```

  

- In file `embot_app_application_theSTRAIN.h` I see an include of a .h file w/ the entire relative path. That compiles and is fine, but it is better to remove the path in inclusions and to add it to the environment variable of the project.

- Your fork is not rebased vs latest `robotology:devel`. However, the changes of the missing PRs are not related to any of the code you have touched and automatic rebase will work fine.

pattacini (changes requested):
>In file embot_app_application_theSTRAIN.h I see an include of a .h file w/ the entire relative path. That compiles and is fine, but it is better to remove the path in inclusions and to add it to the environment variable of the project.

I agree w/ @marcoaccame.
@simeonedussoni @triccyx could you address this, please?

simeonedussoni:
fixed few minor issues:

- version number is now 2.0.13 (note: it is already higher than the IMU remap)
- tabs are replaced with spaces
- included file is without explicit path, which is in the configuration tab

pattacini (approved):
Thanks @triccyx @simeonedussoni 🚀 
Let's proceed with yielding the binaries then!

PR #181, opened 2021-04-21
PMC board: added support for flash pages of 4k

marcoaccame:
This PR adds support for FLASH pages of size 4K to the bootloader and the application of the PMC board.

The modern CAN board, such as the `mtb4` or the `strain2` use FLASH to emulate EEPROM so that they can store CAN address, application version, CAN protocol version and private data. 
They all use a FLASH which is organized in pages of 2K bytes each. There is support to specify a variable inside the `embot::hw::flash` namespace which tells the size of the page, but the functions which manage the FLASH for simplicity have used so far just ... `2048`. 

The first produced batch  of `pmc` boards were programmed with FLASH option bytes which had two banks and page size of 2K. That let them work in the same way as their older sisters.

However, I recently spotted a bug which shows only on the new batch of produced `pmc` boards. The writing to emulated EEPROM did not work.

The reason is due to a different configuration of their FLASH: it had now a single bank and pages of size 4K. This new mode is better (more efficient and more compatible w/ our code) but we need to use the size of the page as a variable.

As a result of that, I have modified some parts of code so that every board can use the proper page size.

For the `pmc` board only I also added some check code inside `bool embot::hw::bsp::specialize()` which executes at startup and performs special operations on the HW.  In here we now detect possible incoherencies between the page size imposed at production and the page size imposed inside our code.

I also moved some FLASH related functions inside namespace `embot::hw::flash` so that we avoid duplications.

I did tests on `pmc` boards of both batches (with pages size of 2K and of 4K) and they all work fine now. The other boards cannot be affected because still keep on using the value 2K as before.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/pmc/bootloader/proj/pmc-bootloader.sct
+++ b/emBODY/eBcode/arch-arm/board/pmc/bootloader/proj/pmc-bootloader.sct
@@ -24,8 +24,8 @@
 ; flash starts from xx and its size is xxx
 ; 80k is: LR_IROM1 0x08000000 0x00014000  {    ; load region size_region
 ; however it was 124K: LR_IROM1 0x08000000 0x0001F000  {    ; load region size_region
-LR_IROM1 0x08000000 0x00013800  {    ; load region size_region
-  ER_IROM1 0x08000000 0x00013800  {  ; load address = execution address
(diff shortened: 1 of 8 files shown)


marcoaccame:
Did further back-compatibility tests on a `mtb4` board.  The EEPROM works fine as expected.

PR #182, opened 2021-04-26
Strain2: remap IMU to stream wrt FT main frame

simeonedussoni:
Modified strain project to have the following axis exchange in the IMU data
- $ x \to y$
- $ y \to -x$
- $ z \to z $

In this way, the FT and the IMU have the same reference frame. 

The code is modified with a conditional preprocessing to insure the compatibility b/w the strain and the other boards equipped with the BNO055, which don't need to have the axis remapped.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/strain2/application/proj/strain2-application-v6.uvprojx
+++ b/emBODY/eBcode/arch-arm/board/strain2/application/proj/strain2-application-v6.uvprojx
@@ -16,8 +16,8 @@
         <TargetCommonOption>
           <Device>STM32L433RCTx</Device>
           <Vendor>STMicroelectronics</Vendor>
-          <PackID>Keil.STM32L4xx_DFP.2.5.0</PackID>
-          <PackURL>https://www.keil.com/pack/</PackURL>
+          <PackID>Keil.STM32L4xx_DFP.2.2.0</PackID>
+          <PackURL>http://www.keil.com/pack</PackURL>
           <Cpu>IRAM(0x20000000,0x0000C000) IRAM2(0x10000000,0x00004000) IROM(0x08000000,0x00040000) CPUTYPE("Cortex-M4") FPU2 DSP CLOCK(12000000) ELITTLE</Cpu>
           <FlashUtilSpec></FlashUtilSpec>
           <StartupFile></StartupFile>
@@ -185,7 +185,6 @@
             <uocXRam>0</uocXRam>
             <RvdsVP>2</RvdsVP>
             <RvdsMve>0</RvdsMve>
-            <RvdsCdeCp>0</RvdsCdeCp>
             <hadIRAM2>1</hadIRAM2>
             <hadIROM2>0</hadIROM2>
             <StupSel>8</StupSel>
@@ -352,7 +351,7 @@
             <NoWarn>0</NoWarn>
             <uSurpInc>0</uSurpInc>
             <useXO>0</useXO>
-            <ClangAsOpt>4</ClangAsOpt>
+            <uClangAs>0</uClangAs>
             <VariousControls>
               <MiscControls></MiscControls>
               <Define></Define>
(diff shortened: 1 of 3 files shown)


pattacini (changes requested):
Thanks @simeonedussoni 

Can you please clean up the PR in order to remove changes to the project files?

Also, please be neat next time w/ the title of the PR 😏

traversaro:
Not critical, but it would be also great to avoid mixing space and tabs, as these results in rendering issues (`#if` and `#else` should be aligned):
tabsandsapces

pattacini:
Good point @traversaro 👍🏻 

>Can you please clean up the PR in order to remove changes to the project files?

Of course, I meant referring only to those files that do not need to be changed as I guess that the `define` has to be saved somewhere in the project.

marcoaccame:
> Not critical, but it would be also great to avoid mixing space and tabs, as these results in rendering issues (#if and #else should be aligned):

yes, I opened the PR this morning because yesterday I spotted that and I wanted address this point.
To avoid presence of tabs it is wise to configure the text editor to substitute the tab w/ 4 spaces.
All (most of) the code in `icub-firmware` and `icub-firmware-shared` is edited w/out tabs because they give a different alignment which depends on the configuration of the editor.

> Can you please clean up the PR in order to remove changes to the project files?

If the project does not change an option such as optimization, include path or similar, we don't need / want to commit its changes.

marcoaccame (changes requested):
Hi @simeonedussoni,

some issue have already been addressed but here they are anyway:
- [ ] no need to commit the project files;
- [ ] change of tabs w/ 4 characters so that the code stays aligned as the style of the existing code;


For the rest, it seems to me that everything is fine. 

I assume that the the rotation is correct and no tests are required.

A comment on the use of macro `STM32HAL_BOARD_STRAIN2`.

The macro `STM32HAL_BOARD_STRAIN2` must be (and is) defined in all the projects running on the strain2 boards. It is designed to offer specialized behaviour to one board and not others. There are similar macros for other boards such as `STM32HAL_BOARD_RFE` etc.

The change you did to file `embot_app_application_theIMU.cpp` must be conditioned under this macro `STM32HAL_BOARD_STRAIN2` because (if) we want to rotate the IMU values only for the `strain2` board but not for the `mtb4` and `rfe` boards which are the other boards using this code.

That is important because the IMU values streamed from the `rfe` are used as they are now for getting the orientation of the head and any change in these values may lead to problems in high level code.

simeonedussoni:
changes requested are implemented.

pattacini (changes requested):
I don't see where the macro `STM32HAL_BOARD_STRAIN2` is defined.
Perhaps in the FT FW project?

My point was to remove changes to those project files that are unrelated but to retain those relevant.

simeonedussoni:
it's defined in the strain2-application-v6 project file, in a tab dedicated to preprocessor macros.

simeonedussoni:
> The macro `STM32HAL_BOARD_STRAIN2` must be (and is) defined in all the projects running on the strain2 boards. It is designed to offer specialized behaviour to one board and not others. There are similar macros for other boards such as `STM32HAL_BOARD_RFE` etc.

as is stated from @marcoaccame here

pattacini:
This file doesn't seem to be contained in this PR though.
So, how could one download the `icub-firmware` repo anew and build the project for the FT FW from scratch?

Am I missing a step?

simeonedussoni:
> 
> 
> This file doesn't seem to be contained in this PR though.
> So, how could one download the `icub-firmware` repo anew and build the project for the FT FW from scratch?
> 
> Am I missing a step?

BTW this is the PR for the IMU remap, after f2f briefing with @marcoaccame we decided to keep the two PR separated. 
In the default project file there is defined among others the preprocessor macro quoted by marco. 
Thus one just needs to download the repo `icub-firmware`, its sibling `icub-firmware-shared` and build the project to obtain a local copy of the executable file. 

As stated above by marco, the project file (which is unchanged from the original one) `strain2-application-v6.uvproj` is not been included in the PR (actually it is just updated by the Keil toolchain when opened with some local environment options which do not impact the binary file generation)

pattacini:
>As stated above by marco, the project file (which is unchanged from the original one) strain2-application-v6.uvproj is not been included in the PR 

So, you're saying that the macro was already defined in the project file? If so, we're ok 👍🏻 
It wasn't clear from the posts above.

pattacini (approved):

pattacini:
Merged ✔
@simeonedussoni could you yield the binaries now?

PR #184, opened 2021-05-19
MBD FOC code optimized with CMSIS-DSP

marcoaccame:
The PR adds a test for MBD (Model Based Designed) code implementing the FOC algorithm (Field Oriented Control).

The FOC code was generated from a Simulink model using some optimization: 
- tuning for Cortex M4 with floating point in single precision,
- use of the CMSIS-DSP library for heavy optimized mathematical operations.

The code was tested on the `amcbldc` board which will run the FOC algorithm inside an IRQ Handler at 80/3 kHz ≈ 26.6 kHz.

But where does the frequency of 26.6 kHz come from? It is the frequency of acquisition of the three currents. The PWM on the `amcbldc` can run at 80 kHz, the three currents are measured in daisy-chain using an ADC converter synchronized with the PWM waveform and are available at the execution of the third IRQ Handler on DMA completion. Hence, at a rate 80/3 kHz ≈ 26.6 kHz.

But how does the code performs on the `amcbldc` running at 160 MHz? 

The tick of the FOC algorithm measures inside a range of [2.5, 4.0] μs depending on optimization. 

|  foc-cmsisdsp-O0   |  foc-cmsisdsp-Ofast  |
| ---- | ---- |

**Figure.** Measure of execution time for the fully optimized MATLAB-generated code: use of single precision, CMSIS-DSP, and compiler `armclang`. Left w/  no optimization (-O0) and right w/ full speed optimization (-Ospeed).

These values surely allow use of the FOC algorithm inside the IRQ Handler.


marcoaccame:
The changes in this PR affects only code used for the `amcbldc` board and do not interfere w/ other code used for the boards on on iCub, so it can be safely merged.

PR #185, opened 2021-06-04
Wrist control integrated to EMS firmware

ale-git:
This PR adds wrist control to the latest devel version of the EMS firmware, with a dedicated project file that defines the WRIST_MK2 compilation macro.


marcoaccame (approved):
ok. let's go

PR #198, opened 2021-07-15
Added Simulink FOC control code for amc-bldc

ale-git:
This PR adds the source code generated by Simulink for motor control (FOC) to the AMC-BLDC board project.
It doesn't control the motor yet, but only includes the required files in the software building.
@marcoaccame already verified that the communication parts is still working fine.


marcoaccame (approved):
ok, @ale-git, that is fine. thank you.

PR #206, opened 2021-10-06
Add AMC-BLDC application04

valegagge:
In this PR we want to add the application04 for AMC-BLDC board. It is tested, please see [here](https://github.com/icub-tech-iit/study-mbd-amc-bldc/issues/52)


marcoaccame (approved):
OK, @valegagge: the PR is fine.

Thank also to @sgiraz @GrmanRodriguez @ale-git @vvasco who worked to achieve this result. 

It is a great work and a great achievement which deserves to be seen in this PR in full.

And in order to show in full the work done in here, I have added some more explanations and the nice video of the demo which runs the motor directly from Matlab.

PR #207, opened 2021-10-08
PMC piezo driver: update wave tables - shrink in range 2V to 44V

andreamerello:
The PMC piezo driver currently uses:
-  precalculated tables to generate the waves required to drive piezo motor phases
-  internally-generated ramps in range `PIEZO_MINVOLTAGE`  to `PIEZO_MAXVOLTAGE` in order to brake and to go in freewheel state (which I don't know whether really works with those motors), and forced `PIEZO_MINVOLTAGE` in case of overcurrent.

Currently the tables were in range 128 to 65407 (DAC codes) while the `PIEZO_MINVOLTAGE` and `PIEZO_MAXVOLTAGE` lead to DAC codes in range  2760 to 60707.

This is clearly visible in this screenshot

pmc

According to G.Z:
- The purpose of clamping the operation range is to avoid driving the voltage outside electronic operating range i.e. otherwise the electronic may clamp the voltage.
- The correct operating range for the electronic is the one specified by  `PIEZO_MINVOLTAGE`  and `PIEZO_MAXVOLTAGE` constants

This lead to the conclusion that the wave tables should be recalculated to fit in the 2760 to 60707 range.

This PR updateds the tables and the generation script according to this. 

The outcome is visible in the following screenshot. NOTE: It has been not tested on real HW.

pmc2


marcoaccame (dismissed):
Hi @andreamerello: thanks very much for this change. 

I believe you analysis is right. On the other hand you and G.Z. (aka @giorgiozini) produced the original tables for this PZM driver.

We shall test the new tables on the pmc board (@simeonedussoni can you pls do that and report it on the PR?) and then we merge the code.

marcoaccame (approved):
Hi @andreamerello thanks for your contribution. The changes you propose are fine and shall be merged. Sorry if we waited so long for this to happen.

PR #215, opened 2021-11-10
bmon firmware added

MrAndrea:
This PR to add the `BMON` electronic board firmware in the `icub-firmware` repository.

The `BMON` board is the small board mounted on the iCub 2.5 battery; it has 4 LEDs and it communicates with the BCB board through I2C bus; it embeds a Microchip PIC16F1824 microcontroller. I (@MrAndrea ) wrote the firmware many years ago. It was inside the electronics repo and it is worth to move it in here.


marcoaccame:
Hi @MrAndrea, it is better to do the PR vs the `devel` branch. At end of November it will be in `master` anyway because we are planning a new release. 

I will do the change from `master` to `devel` myself.

MrAndrea:
> Hi @MrAndrea, it is better to do the PR vs the `devel` branch. At end of November it will be in `master` anyway because we are planning a new release.
> I will do the change from `master` to `devel` myself.

Ok, thank you @marcoaccame.

marcoaccame (approved):
Hi @MrAndrea, the PR can be safely merged as it does not touch existing code.

PR #217, opened 2021-11-12
No nonius management of AMO sensors

ale-git:
Now, AMO sensors can be read as multi-sector (64 or 32) 14-bit resolution encoders (20 or 19-bit resolution total).

The calibration is hard-stop (type 10) as in R1 wrist pronosupination.


marcoaccame (commented):
Hi @ale-git, I see that in `eo_appEncReader_GetValue()` you now filter differently the `spiRawValue` according to the number of sectors. I also see that you use only only 14 bits.
**Question**: is this change compatible w/ the absolute reading of AMO which is still used in upper body of iCubGenova09 w/ calibration type 12, so that we can use the same code for upper and lower body?

marcoaccame (changes requested):
Also, we need to advance the application version of the boards affected by the changes. It is only the `ems`, correct?

ale-git:
Hi @marcoaccame, yes, it is fully back compatible with present management of AMO and AEA boards used as absolute encoders with calibration type 12 and GearboxE2J = 1 parameter. It only affects the EMS unless we have AMO boards connected with MC4+ board (we have no one at present).

ale-git:
But, in case of AMO connected to MC4+ boards it should be compatible as well.

marcoaccame (approved):

PR #226, opened 2021-12-15
AMC_BLDC: Alignment with the last changes on MDB

sgiraz:
What's changed in this PR:
- In application05:
  - embot_app_application_theMBDagent.cpp contains a cleaner code
  - the code generated from the MBD (master) now contains the last changes. The supervisor now changes control mode and manages/prevents HF correctly.
  - embot_tools has been introduced for performance analysis
- In application04:
  - this folder contains the FW version before the refactoring of the mbd
  - fixed the change of the control. Now it is performed correctly by the Supervisor.

**Note for application05:**
- In this PR, the problem of motor control is still in progress. Targets of current values greater than ~300mA will trigger the Supervisor HF-Handler. The fault triggered by the Supervisor is correct (it isn't a bug). The problem is in the control.

cc @valegagge @GrmanRodriguez @vvasco @ale-git


marcoaccame (approved):
we can safely merge because this PR does not affect any robot on the robot (yet).

PR #227, opened 2021-12-16
Mutual exclusion in amcbldc

marcoaccame:
This PR just adds a mechanism of mutual exclusion in the `amcbldc` board between a `user thread` and `DMA1_Channel2_IRQHandler(`) which both run MBD code.

The code was successfully tested on the `amcbldc` board. 

Moreover, this change has a local impact only on the `amcbldc` and hence it can be safely merged.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amcbldc/application05/proj/amcbldc-application05.uvoptx
+++ b/emBODY/eBcode/arch-arm/board/amcbldc/application05/proj/amcbldc-application05.uvoptx
@@ -125,7 +125,7 @@
         <SetRegEntry>
           <Number>0</Number>
           <Key>ULP2CM3</Key>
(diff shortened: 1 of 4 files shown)


pattacini (commented):
Hi @marcoaccame 

Do you think we could define the implementation of lock and unlock as inline functions to spare the context switch?

marcoaccame:
> Hi @marcoaccame
> 
> Do you think we could define the implementation of lock and unlock as inline functions to spare the context switch?

I don't think it changes much to avoid a function call @ 160 MHz on a CM4, but it is worth having the implementation. See https://github.com/robotology/icub-firmware/pull/227/commits/94761d43421d9611a36ff0ae32afd2c703ba1cb5.

pattacini:
Regarding `RTW_MUTEX_USE_INLINE`, maybe it's just for the test, but I deem it's a complication that we can spare.

PR #229, opened 2021-12-20
Cleanup of amcbldc projects

marcoaccame:
This PR contains a cleanup of the `amcbldc` projects from `application01` to `application05`.

These are the projects that we have prepared to test the motion control based on MBDE on the `amcbldc`.

The projects are developed in a evolutive way and we add more and more features as long as the development goes on.
However, in this evolution the older projects did not compile anymore and the latter two, `application04` and `application05`, required some refactoring and also tools able to measure the execution time of the two processing loops: the FOC inner loop @ 37.5 us and the tick loop at 1 ms.

This PR contains such changes.

We have verified that the functionalities have stayed unchanged, so we can safely merge this changes.


emBODY/eBcode/arch-arm/board/amcbldc/application04/src/embot_app_application_theMBDagent.cpp
```diff
@@ -304,19 +364,129 @@ bool embot::app::application::theMBDagent::Impl::tick(const std::vector<embot::p
...
+//        FOC_inputs* u = (FOC_inputs*)rtu; -> becomes impl->foc_inputs.
+//        FOC_outputs* y = (FOC_outputs*)rty; -> becomes impl->foc_outputs.
+        
+        embot::hw::motor::gethallstatus(embot::hw::MOTOR::one, impl->foc_inputs.sensors_data.motorsensors.hallABC);
+        
+        embot::hw::motor::Position electricalAngle;
+        embot::hw::motor::getencoder(embot::hw::MOTOR::one, electricalAngle); 
+        
+        static int32_T position;
+        static int32_T electricalAngleOld =  static_cast<real32_T>(electricalAngle);
```
sgiraz, line 401: Hi @marcoaccame, In here we should replace `realt32_T` with `int32_T` because `Position` is an `int32_t` type.

PR #235, opened 2022-01-10
AMC_BLDC: Update mbd generated code with the last changes that fix the FOCcontrol issue

sgiraz:
What's changed in this PR:

- Finally, the problem of motor control has been solved. So, the generated code has been updated with the last changes of the mbd. 
- The commands to enable/disable the motor are now part of the mbd.
- In `rtw_mutex.h` the mutex functions are now always enabled.

**Note:** 
When we send the current values for the target, we should pay attention to the current limits we set before.
Even limits of 772mA with a target of 470mA can trigger the Supervisor Fault Handler.
This behavior is not an issue of application05 because it happens in application04 as well.

cc @valegagge @GrmanRodriguez @vvasco @ale-git


pattacini:
Great 🚀 

>Even limits of 772mA with a target of 4760mA can trigger the Supervisor Fault Handler.

The overloadcurrent limit is meant to be quite high ~ 20 A; hence, it ain't a problem, definitely.

The peakcurrent limit, instead, will be used to implement the SW current limiter, which is now missing.

marcoaccame (approved):
Hi @sgiraz. That is all fine, thanks.

PR #237, opened 2022-01-11
AMCBLDC: get CAN address from the cached CANBoardInfo

valegagge:
In this PR the firmware uses the CAN address stored in ROM instead of the constant fixed in the model (its value is 3)


marcoaccame (approved):
well done!

PR #257, opened 2022-03-14
FT service with multiple sensors

marcoaccame:
### Description
This PR with its companion PRs in [icub-firmware-shared](https://github.com/robotology/icub-firmware-shared/pull/58) and [icub-firmware-build](https://github.com/robotology/icub-firmware-build/pull/52) allow an ETH board such as the `ems`, `mc4plus`, `mc2plus` to stream FT data and associated temperature coming from one, two or more CAN FT sensor boards such as the `strain2` or the `strain`.

The changes generated by the above triad of PRs (`icub-firmware`, `icub-firmware-shared`, `icub-firmware-build`) can coexist safely with current release of `icub-main/devel`, even if the multiple FT service cannot be activated yet.

In order to see the multiple streams at YARP level, we need to add in `icub-main` the device `embObjMultipleFTsensors` with a PR which will be soon made available.

So far, I have done extensive tests on a dedicated setup to verify the following.
- **backwards compatibility**. `yarprobotinterface` if compiled w/ icub-firmware-shared of the companion [PR](https://github.com/robotology/icub-firmware-shared/pull/58) can safely run:
   -  a standard FT service on an `ems` board,
   -  a MC service on an `mc4plus` board driving a single motor.
- **capability of streaming from multiple sensor boards**.  An `ems` board with two `strain2` attached has run a simulation where the object `embot::app::eth::theFTserviceTester` was faking the presence of `embObjMultipleFTsensors`. The result was that the new object `embot::app::eth::theFTservice` was correctly activated and deactivated as if `yarprobotinterface` were present.
- **verification of transmission from the sensor boards**.  The `embot::app::eth::theFTservice` object runs also a new `embot::app::eth::CANmonitor` service which is able to check if the required CAN boards transmits regularly. The `CANmonitor` can be configured to check at a given rate and to send diagnostics messages in various modes, from minimal to fully pedantic.

   
### More details

Here are the more details on how the `embot::app::eth::theFTservice` can be configured at xml level and what FT sensors data sends up to yarprobotinterface.

```xml
    <group name="SERVICE">
        
        <param name="type"> eomn_serv_AS_ft </param>
    
        <group name="PROPERTIES">

            <!-- we can have as many boards as we want, 
                 but now the only good ones are strain and strain2. 
                 use eoas_ft_isboardvalid() to verify -->
            <group name="CANBOARDS">
                <param name="type">                 strain2     strain      </param>

                <group name="PROTOCOL">
                    <param name="major">            2           1           </param>    
                    <param name="minor">            0           0           </param>     
                </group>                    
                <group name="FIRMWARE">
                    <param name="major">            2           1           </param>    
                    <param name="minor">            0           1           </param> 
                    <param name="build">            9           3           </param>
                </group>
            </group>
            
            <!-- we can have as many sensors as we want. however ... -->
            <group name="SENSORS">
                <param name="id">                   l_foot_ft1      l_foot_ft2      l_foot_ft3  </param>
                <param name="board">                strain2         strain2         strain      </param>
                <param name="location">             CAN2:13         CAN1:12         CAN2:11     </param>
            </group>                
        
        </group>

        <!-- we can pick up at max 4 sensors from the above list. 
             we can specify parameters which are different for each sensor 
             - ftPeriod is expressed in ms in range [1, 250]
             - temperaturePeriod is expressed in ms (but it will be used in seconds).
             - useCalibration is boolean
             -->
        <group name="SETTINGS">        
            <param name="enabledSensors">       l_foot_ft1  l_foot_ft2  l_foot_ft3      </param>
            <param name="ftPeriod">             10          10          10              </param>
            <param name="temperaturePeriod">    1000        1000        0               </param>
            <param name="useCalibration">       true        false       true            </param>           
        </group>       
     
        <group name="CANMONITOR">        
            <param name="checkPeriod">      100     </param>
            <param name="reportMode">       ALL     </param>
            <param name="ratePeriod">       20*1000 </param>
        </group>    
     
    </group>
```
**Code Listing**. Configuration of the multiple FT service. In this case, we ask FT data from three sensors each with a period of 10 ms (but we could also specify a different period for each). Two sensors stream calibrated data, one uncalibrated. Two sensors also stream temperature at 1 second period. 

```xml
<!-- CANMONITOR tells the ETH board how to verify the continuous presence of the
     CAN boards which are required for the service and how to report to
     yarprobotinterface w/ suitable diagnostics messages.
     Here are the parameters.

     - checkPeriod: in milliseconds in range [0, 255]. It is the period at which the 
       ETH board monitors if all CAN boards are present in the bus. 
       If 0 there is no monitoring
     - reportMode: it tells how the ETH board sends up diagnostics messages on
       the result of its monitoring. The following string values are allowed:
       - NEVER: the ETH board never sends a diagnostic message about the monitoring status)
       - justLOSTjustFOUND: sends diagnostics only when boards disappears and when
         they reappers.
       - justLOSTjustFOUNDstillLOST: as justLOSTjustFOUND but the ETH board also sends 
         periodic diagnostics if boards are still found but with a rate of 
         reportPeriod millisec.
       - ALL: as justLOSTjustFOUNDstillLOST but the ETH board also sends periodic diagnostics
         even if all boards are present.
     - reportPeriod: the period in milliseconds at which the ETH board sends diagnostics messages
       in case the CAN boards are still missing or even if they are present. This
       value has meaning only if reportmode is justLOSTjustFOUNDstillLOST or
       justLOSTjustFOUNDstillLOST. Values to use are in range [1000, 64*1024-1] 
  -->
<group name="CANMONITOR">        
    <param name="checkPeriod">      100     </param>
    <param name="reportMode">       ALL     </param>
    <param name="ratePeriod">       20*1000 </param>
</group>  
```
**Code Listing**. Configuration of the CAN monitor. In this case, we check vs presence of the CAN boards every 100 ms, we report with diagnostics messages every 20 sec if boards are all OK, with one immediate message if any goes away or if it returns and every 20 sec if they should not be back.

```C++
typedef struct
{
    // timeoflife in usec. better using this because yarp may ask the board to have its time
    eOabstime_t             age;      
    // ffu ... used-calib-set, saturated info, calib/notcalib, candata-not-heard-of-since-ms-ago, 
    uint32_t                info;         
    uint16_t                ffu;
    // in steps of 0.1 celsius degree (pos and neg).
    eOmeas_temperature_t    temperature; 
    // they may be [f, t] or also the 6 adc values or also ... see info.     
    float32_t               values[6];       
} eOas_ft_timedvalue_t;     EO_VERIFYsizeof(eOas_ft_timedvalue_t, 40)
```
**Code Listing**. The FT service streams up to `yarprobotinterface` the 40 bytes of `eOas_ft_timedvalue_t` for each FT sensor. The six FT values are in floating point format so that we are ready for future better precision in the CAN sensor. They can contain either calibrated or raw values.  We also have the temperature and the acquisition time of the FT values. There are also several bits which may be useful later on for extra signalling / diagnostics (i.e., for indicating presence of saturation, type of calibration used etc.). 

### What's next
As a result of the **backwards compatibility** tests reported above, I believe that these three PRs can be safely merged. When the new PR on `icub-main` is ready we can perform more tests on dedicated setup and on the robot.


emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theFTservice.cpp
```diff
@@ -0,0 +1,1912 @@
...
+}
+
+   
+eOresult_t embot::app::eth::theFTservice::Impl::Start()
+{
+    eo_errman_Trace(eo_errman_GetHandle(), "::Start()", s_eobj_ownname);
+    
+    if(eobool_false == service.active)
+    {   // nothing to do because object must be first activated
+        return(eores_OK);
```
sgiraz, line 735: Hi @marcoaccame, probably it will not change much, but would it be better to return `eores_NOK_generic` instead of `eores_OK`?

PR #271, opened 2022-05-26
Enhancement of the `embot::hw::timer` driver

marcoaccame:
This PR enhances the `embot::hw::timer` driver in two ways:
- it adds some extra functions required for the porting of the objects `EOMtheRunner` to the `amc` application,
- it adds the support of emulated HW timers.

The emulation is implemented w/ the `embot::os::Timer` objects and is enabled w/ macros `EMBOT_ENABLE_hw_timer` and `EMBOT_ENABLE_hw_timer_emulated`.

This feature is useful because I can begin testing the control loop inside the `EOMtheRunner` using some sort of timers which can still do the job, even if not surely at 1KHz.

In the meantime the development of the `embot::hw::timer` using the HW of the new H7 CPU can proceed, hopefully in parallel.

This change does not affect other boards which use the same `embot::hw::timer` driver because I have added functions, and not changed or removed existing ones.

So, this PR can be safely merged.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/bsp/embot_hw_bsp_amc.cpp
+++ b/emBODY/eBcode/arch-arm/board/amc/bsp/embot_hw_bsp_amc.cpp
@@ -1322,5 +1322,62 @@ extern "C"
 // - support map: end of embot::hw::spi
 
 
+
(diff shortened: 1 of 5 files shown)


marcoaccame:
cc @sgiraz

PR #272, opened 2022-05-30
Stream of battery data over CAN: documented the protocol

marcoaccame:
TSD-ICUBUNIT-canprotocol-sensorboards.docx: added a class and a message able to stream battery data.


marcoaccame:
cc @triccyx

PR #273, opened 2022-06-08
AMC_BLDC: Clean CAN decoder, Fix Volt units, add SW current limiter in Volt mode

mfussi66:
This PR marks the first code generation with MATLAB R2022a out of PR 218 of the architectural model.

Highlights:

- AMC_BLDC: Clean up the CAN Decoder
- AMC_BLDC: Fix voltage units
- AMC_BLDC: Implement SW Current Limiter in Voltage Control Mode
- AMC_BLDC: addition of `filter-current` -> resulted in addition of include path and group in Keil project file


emBODY/eBcode/arch-arm/board/amcbldc/application07/src/embot_app_application_theMBDagent.cpp
```diff
@@ -392,7 +411,22 @@ void embot::app::application::theMBDagent::Impl::onCurrents_FOC_innerloop(void *
     {
         return;
     }    
-        
+    
+    // performance measurement
+    if(true != useDUMMYforFOC)
```
pattacini, line 416:
```suggestion
    if(false == useDUMMYforFOC) 
```

Just to equalize the condition and make it the same as the one above

pattacini (approved):
I was about to merge the PR thinking that it touched only the codegen, but then I saw the changes to the module `embot_app_application_theMBDagent.cpp`.

Those changes look to pertain to some debugging code though; hence all good.
But maybe, maintainers want to chime in here.

cc @marcoaccame

marcoaccame:
> Those changes look to pertain to some debugging code though; hence all good.
> But maybe, maintainers want to chime in here.

Yes @pattacini thanks. That is code which I asked @mfussi66 to add in the PR to better supprt the profiling of the execution time. I will review it quite soon.

valegagge:
> Those changes look to pertain to some debugging code though; hence all good.
But maybe, maintainers want to chime in here.

I already tested this code as you can see [here](https://github.com/icub-tech-iit/study-mbd-amc-bldc/issues/207). This code is helpful for testing the performance, even if it should be better to improve it.

@mfussi66 Why did you commit the project file `amcbldc-application07.uvprojx`? Did you change some project configuration?

pattacini:
>I already tested this code as you can see https://github.com/icub-tech-iit/study-mbd-amc-bldc/issues/207. This code is helpful for testing the performance, even if it should be better to improve it.

👍🏻 

>@mfussi66 Why did you commit the project file amcbldc-application07.uvprojx? Did you change some project configuration?

I think the following holds from the OP:

>AMC_BLDC: addition of filter-current -> resulted in addition of include path and group in Keil project file

marcoaccame:
> I already tested this code as you can see [here](https://github.com/icub-tech-iit/study-mbd-amc-bldc/issues/207). This code is helpful for testing the performance, even if it should be better to improve it.

@valegagge, I added the improvements you suggested such as the name of the owner of `MeasureHisto`.

pattacini:
Hi @marcoaccame @valegagge 

Can we merge the PR so that we can proceed with the second one?

marcoaccame:
> Hi @marcoaccame @valegagge
> 
> Can we merge the PR so that we can proceed with the second one?

ok for me

PR #276, opened 2022-06-15
Application 07: Set FOC and STATUS messages as response to DESIRED_CURRENT message

GrmanRodriguez:
This PR updates Application 07 to the latest version of the AMC-BLDC model. Currently, that is [this](https://github.com/icub-tech-iit/study-mbd-amc-bldc/commit/305b7ac313fd6443271adcffc3632089a4475dd8) commit.

This includes the change in [this](https://github.com/icub-tech-iit/study-mbd-amc-bldc/pull/220) PR, which makes the board send the `FOC` and `STATUS` messages only as a response to a `SET_DESIRED_CURRENT` message. Before, the messages would be sent continuously without ever stopping after the first desired current request.


pattacini (approved):
Only `codegen` is touched.
I think we can merge the PR straight away.

cc @marcoaccame @mfussi66

PR #283, opened 2022-07-14
Application07: Update generated code to support new messages

GrmanRodriguez:
This PR updates the generated code of Application 07 of AMC BLDC to the currently latest commit of [`b0dd4901a`](https://github.com/icub-tech-iit/study-mbd-amc-bldc/commit/b0dd4901a20cb94d2b567abaac110106ae0a1cc1).

Specifically this means the board is now able to process the new set motor configuration, set velocity PID, set current PID and set desired target messages.

This has been tested on hardware and the setup works as well as the processing of the new messages.


marcoaccame (approved):
ok, thanks

PR #294, opened 2022-09-01
AMCBLDC: latest codegen: Improve units handling for velocity PID parameters

mfussi66:
This PR updates the generated code for the AMCBLDC board, coming from `study-mbd-amcbldc@3c50604235`. 
Highlight:
- The CAN decoder can handle Velocity PID parameters with a unit defined as mA / (deg/s)
- Sensible velocity PID gains are set up by default to achieve steady state in ~1.5sec


mfussi66:
cc @pattacini @sgiraz @marcoaccame

pattacini (approved):

PR #316, opened 2022-10-28
hal-spiencoder: small fix in hal_spiencoder_get_value2 for aksim2 encoder

valegagge:
Here I'm pushing a minimal fix in the hal_spiencoder_get_value2, which I already tested with @sgiraz.

In this way, we can check all status-bits.


sgiraz (approved):
Ok @valegagge, thanks!

PR #332, opened 2023-01-11
AMCBLDC: Fix computation of angle when using quadrature encoder

mfussi66:
This PR is aimed at the AMCBLDC Hal, to fix:
### Detection of the sector of the phases 
`line 231 in pwm.c`

In this line we compute the sector that the electric angle belongs to = [0, 1, ..., 5].
Casting to int16 would cause sector to have the values [-5 -3 -1 1 3 5], which would mess up the later computations

### Desynching between hall sensors angle and quadrature encoder angle
A reset function of the quadrature encoder angle is applied whenever a full rotation is done.
A full rotation is defined as the transition between sector 5 and sector 0.
This procedure is also used to during calibration step 0, to transition to step 1:
- Step 0:  Rotate using the Hall sensors until there's the switch between sector 5 and 0, reset the encoder, and go to step 1
- Step 1: Rotate using the Hall sensors until we read the uncalibrated encoder value for each of the 6 sectors (or borders);
Then, we compute the offset by taking the average of the 6 values, and apply it.
- Step 2: Store the electric angle evaluated by the hall sensors as always, and reset the encoder if a full rotation is done

cc @sgiraz @pattacini @valegagge

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amcbldc/application/src/motorhal/encoder.c
+++ b/emBODY/eBcode/arch-arm/board/amcbldc/application/src/motorhal/encoder.c
@@ -200,7 +200,9 @@ uint32_t encoderGetCounter(void)
     return (uint32_t)__HAL_TIM_GET_COUNTER(&htim2);
 }
 
-
+/*******************************************************************************************************************//**
+ * @brief   Reset encoder value
+ */
 void encoderReset()
 {
     __HAL_TIM_SET_COUNTER(&htim2, 0);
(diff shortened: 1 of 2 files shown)


marcoaccame (changes requested):
Hi @mfussi66, the change is fine. I assume that it was tested on a dedicated and everything works.

It is important however also to explain how are obtained the magic numbers used in `pwm.c`.

It may be enough a brief description inside the PR which points to an inline comment in the code w/ more details. 

Later on (a further PR is fine) we shall collect all the comments inside a `motorhal/docs/readme.md` file which documents the internals of the motorhal.

mfussi66:
@marcoaccame I added comments to the code and some constexpr to give meaning to the hardcoded numbers, let me know if they add clarity to the code.
I'll keep testing the motor to make sure it all works as expected before marking the pr ready for review.

marcoaccame (approved):
Ok, thanks

PR #339, opened 2023-01-25
The POS service offered by `EOthePOS` can share the CAN bus used by another ETH board

marcoaccame:
This PR adds the required protections so that the POS service can be run by multiple ETH boards which share the same bus as long as they use CAN boards w/ different ID.

To clarify:

If board `ETH-1` configures a CAN board `can-ID-1` to transmit POS values inside a CAN frame of class streaming analog, these messages are received by any ETH board attached to the same CAN bus. Suppose for instance that the board `ETH-2` shares the same bus.

Both `ETH-1` and `ETH-2` will receive the same messages. `ETH-1` expects and correctly processes them, but `ETH-2`does not expect them and must discard them.

This PR adds a filter to the CAN RX processing functions inside `EOtheCANdiscovery2` and `EOthePOS` so that they process only the messages coming from the CAN IDs used by the service.

This PR is required because in the new hand V3 we use two mc4plus boards which share the same CAN bus and both of them run the POS service using two different `mtb4c` boards.


marcoaccame:
tested on the robot and it works.

PR #342, opened 2023-02-17
rfe test firmware

davidetome:
This PR adds the `rfe` test firmware.

Some tests will be validated once external cables will be ready.

cc @marcoaccame @pattacini @maggia80 @sgiraz @MrAndrea


marcoaccame:
Hi @davidetome, I will have a look at that tomorrow morning.

emBODY/eBcode/arch-arm/board/rfe/bsp/embot_hw_bsp_rfe.cpp
```diff
@@ -605,10 +605,10 @@ namespace embot { namespace hw { namespace i2c {
         {            
             MX_I2C1_Init();
         }
-//        else if(h == I2C::two)
```
marcoaccame, line 608: Hi @davidetome, we don't need this initialization. It never runs because `I2C::two` is not enabled in the `embot::hw::i2c::thebsp`.

emBODY/eBcode/arch-arm/board/rfe/test/src/STM32_USB_Device_Library/Class/CDC/Inc/usbd_cdc.h
```diff
@@ -0,0 +1,205 @@
+/**
```
marcoaccame, line 1: The folders `STM32_USB_Device_Library` and  `usb_user` are already under `rfe/application/src`. If the code hasn't changed it is better to include them in the `rfe-test-v6` project rather than copying them as they are below `test/src`

marcoaccame (changes requested):
Hi @davidetome, the PR is fine but pls have a look at the comments. If the USB driver is unchanged it is better to remove the duplicated files.

PR #353, opened 2023-03-03
BAT board: align the software overcurrent thresholds to the hardware thresholds

MrAndrea:
This PR is to align the software overcurrent thresholds of the BAT board to the corresponding new hardware revision: iitcode 10050 rev. D.
Note: all the previous revisions are obsolete.

The hardware overcurrent limits are:
- 36A for the HSM section
- 10A for the 12V voltage regulator CPU section
- 10A for the 12V voltage regulator Motor section

PR #354, opened 2023-03-03
BAT firmware revision version updated to 1.1.0

MrAndrea:
I forgot to update the revision version in https://github.com/robotology/icub-firmware/pull/353.
This PR is only to change it.


marcoaccame (approved):
OK. 

We can merge it into `master` as previous [PR](https://github.com/robotology/icub-firmware/pull/353) did the same. And after we shall merge `master `into `devel`. Let's say that it is a bugfix.

PR #355, opened 2023-03-07
The board `amcbldc` upgrades its low level HW motor driver

marcoaccame:
This PR upgrades the `amcbldc` project to use a new low level HW motor driver, the so called `motorhal`.

The `motorhal` is code written on top of the ST HAL library which configures and uses HW peripherals such ADC and TIM to read currents, voltages and to produce the PWM for the motor.

The new `motorhal2` has been improved over the legacy `motorhal` to use a new version of ST HAL and some improved configurations of the HW peripherals.

At the same time the legacy driver has been mildly refactored.

The new project can use either the legacy or the new driver to enable comparative tests.

In the following figure is show what has changed.

```
   ---------------------------
  | model based motor control | <- unchanged
   ---------------------------
  | embot::hw::motor          | <- almost unchanged
   ---------------------------  
   ---------------------------
  | motorhal                  | <- either motorhal or motorhal2
   ---------------------------  
   ---------------------------
  | STM HAL                   | <- either v120 or v122
   --------------------------- 
   
``` 
**Figure**. Changes in the motor control stack in `amcbldc`.

## Tests

Test were performed and showed similar behaviour w/ the previous HW motor drivers. So we can safely merge. In any case, I have kept the previous drivers as an alternative target in the `amcbldc` project.


sgiraz:
@marcoaccame and I tested successfully both the intermediate `v1.1.10` and the definitive `v1.2.10`  FWs on the `wrist-setup`

cc @Nicogene @pattacini

PR #349, opened 2023-02-28
Improved calib type 6 w/ FAP.

MSECode:
Tested and working on both desk setup and ergocab robot.

In this PR the following tasks have been done:
- Improve current working calibration type 6 by allowing to use FAP sensors instead of MICE;
- Tested the calibration procedure on the left arm bench setup.
- Tested the calibration procedure on the left arm of ergocub robot (test on right arm is going to be done)
- Updated firmware code by customizing calibration type 6 enabling POS service use

PR #356, opened 2023-03-08
Fix of PR #349 for problem related to POS service

MSECode:
This PR introduces a dummy function for the method `eo_pos_isAlive` when the POS service is disabled in a board.
This has been done in order to fix compilation problem for projects that do not used POS service introduced with the PR #349 where a file shared among projects has been modified.


MSECode:
cc: @valegagge @sgiraz @marcoaccame

marcoaccame (approved):
OK, thanks

PR #357, opened 2023-03-09
Ergojoint support added (under #ifdef ERGOJOINT)

ale-git:
I've added support to torque measure in ergoCub new joint based on harmonic drive torsion.
To enable this feature ERGOJOINT macro must be defined in the project configuration.
The project built without ERGOJOINT macro is full compatible with standard EMS fw.


maggia80 (commented):
@ale-git could you open a PR in icub-firmware-models where you define the original Simulink model from which you generated the code?

marcoaccame (changes requested):
Hi, here are a few changes which may help:

- rebase vs latest `devel`  and solve conflicts. 
- create a PR of the used torque model inside https://github.com/robotology/icub-firmware-models and  make a reference to it inside this PR.
- move the model-based generated code under folder `emBODY\eBcode\arch-arm\mbd` as for instance done w/ the `kalman_filter`. 

Then, I assume that this PR will keep an experimental FW not yet to be deployed inside the robot. It will be in future, right? If so:

- revert the project `ems4rd.diagnostic2ready.uvprojx`so that is not affected (yet) by the new code. I see that it already compiles the new files.

- I suggest to shape the version number of the `ems4rd.ergojoint.uvprojx` using macro `ERGOJOINT`. That makes the experimental FW recognizable and at the same time it avoids conflicts in file `EOMtheEMSapplCfg_cfg.h`. Something like that:

  ```c++
  #if defined(WRIST_MK2)
      #define VERSION_MAJOR_OFFSET  20
  #elif defined(ERGOJOINT)
      #define VERSION_MAJOR_OFFSET 30
  #else
      #define VERSION_MAJOR_OFFSET 0
  #endif
  ```

emBODY/eBcode/arch-arm/embobj/plus/board/EOappEncodersReader.c
```diff
@@ -647,11 +648,18 @@ extern eOresult_t eo_appEncReader_GetValue(EOappEncReader *p, uint8_t jomo, eOen
                     // check validity for aksim2
                     if(eobool_true == s_eo_appEncReader_IsValidValue_AKSIM2(&diagn, &prop.valueinfo->errortype))
                     {
+                        #ifdef ERGOJOINT
```
marcoaccame, line 651: One more thing: would it be possible to call the `MController_motor_calc_torque*()` functions inside the `MController`, for instance inside `MController_update_absEncoder_fbk()` which is the one that takes care of the encoder values? 

The benefit in this is that we keep the code more encapsulated. And also, we can have the torque computation for free also in the `amc` and future dual core boards which use a new simplified [encoder-reader](https://github.com/robotology/icub-firmware/blob/master/emBODY/eBcode/arch-arm/embot/app/eth/embot_app_eth_theEncoderReader.cpp).

valegagge:
> @ale-git could you open a PR in icub-firmware-models where you define the original Simulink model from which you generated the code?

It is in [ergocub-design-joint/blob/sim/ErgoJointTorqueControlGen.slx](https://github.com/icub-tech-iit/ergocub-design-joint/blob/sim/ErgoJointTorqueControlGen.slx)

ale-git:
We decided to suspend this PR, so I'm gong to put it in "on hold" state.

marcoaccame:
Hi @ale-git, I propose to close this PR as it is quite old and I believe it is maybe too difficult to rebase the fork vs latest devel. 
You can still re-open another one w/ similar content if you need.

marcoaccame:
> Hi @ale-git, I propose to close this PR as it is quite old and I believe it is maybe too difficult to rebase the fork vs latest devel. You can still re-open another one w/ similar content if you need.

OK. I will close then

PR #358, opened 2023-03-13
amcbldc: Check overload current limit against filtered Iq

mfussi66:
This PR aligns the amcbldc codegen to the latest development in icub-firmware-models: https://github.com/robotology/icub-firmware-models/pull/21#issuecomment-1466180180

This change allows greater tolerance against the very noisy Iq current, otherwise an overcurrent hardware fault would be triggered too easily.


sgiraz:
> This change allows greater tolerance against the very noisy Iq current, otherwise an overcurrent hardware fault would be triggered too easily.

cc @maggia80

sgiraz (approved):
Ok @mfussi66, thanks!

PR #376, opened 2023-05-03
Layer `embot::hw::motor` + `motorhal` for board `amc2c`

marcoaccame:
This PR adds in `embot::hw::motor`  for the `amc2c` board all that is required to drive the motor with the MBD code.

This code guarantees the legacy HW tests of previous PR.

Some extra basic tests are done (e.g., the calibration mechanism of the ADC acquisition) so we are ready for the tests on the motor w/ MBD.


sgiraz:
Hi @marcoaccame, I'm going to this test right now.

PR #378, opened 2023-05-11
Added Hall sensor effect value and sequence error detection.

ale-git:
Three functions are added to the HAL to check the error state related to Hall sensors and overcurrent - the latter was already monitored but not reported.

It just adds functions to check the error state, but takes no action, that is demanded to the supervisor, so it should be safe to merge.


marcoaccame:
Hi @ale-git, I will soon look at the PR. Pls hold on.

marcoaccame (approved):
Hi @ale-git, the PR is fine thanks.

PR #382, opened 2023-05-22
2FOC v3.6

ale-git:
Since this version of the 2FOC firmware introduces some radical improvements, its project is added to the repository and intended to cohexist with the V3 version for some time, until well tested.

Main features are:

-  Quadrature encoders not limited to 16 bit resolution
- Smoother integration between HALL sensors and quadrature encoders
- Fake warnings removed, better heuristics
- Improved current sensing with bias and gain calibration at startup
- Feedback current filtering


ale-git:
I've tested this version with the new ErgoCub joint (Hall sensorless motor) as well as with an old 2.x iCub motor (Hall + encoder motor, without index).

marcoaccame (approved):
Ok @ale-git, the code of the PR is fine. So I will merge it. Only thing as we said f2f pls feel free to add some comments in the PR itself about where to pick up the new 2FOC project.

ale-git:
### Main differences in encoder management
The `POSCNT` encoder count register in 2FOC dspic33f processor is limited to 16 bit, and this limitation affects older versions of the 2FOC firmware, but not this last. The encoder resolution configuration parameter in CAN configuration protocol is intepreted in this new version not as CPR but as equivalent encoder ticks per revoution from the sensor system, **before** considering that the 2FOC internally counts each raise/fall front, ths multiplying by 4. Example: if the encoder wheel has 1300 ticks, and the sensor board performs a 16x interpolation, the encoder resolution written into the configuration file will be 16*1300 = 20800. The internal CPR in the 2FOC board will be 4*20800 = 83200 CPR. In this way the resolution we can write in the configuration file and transmit it over CAN is +- 131072 instead of previous +- 32768.
NOTICE: the configuration files for old robots must be updated accordingly, with resolution divided by 4. For example, in iCub 2.x robots the encoders have 900*4*4 = 14400 cpr, that must be changed to 900*4 = 3600 cpr.

ale-git:
### Internal encoder management
The BLDC motors usually have several polar pairs. For us, they ranges from 4 (iCub 2.x and iCub V3), 6 (iCub V3) and 10 (Ergocub).

- Without Hall sensors

In the new firmware the encoder `MAXCOUNT` is set to the range of an _electrical revolution_, that corresponds to a mechanical revolution divided by the number of polar pairs. Example: with 10 polar pairs, the encoder is reset every 360 / 10 mechanical degrees. In this way the 16 bits of the` POSCOUNT` register is sufficient to support high resolution quadrature encoders. Moreover, the internal management is very simplified in order to perform index synchronization.

- With Hall sensors

The electrical revolution of the motor is subdivided into 6 sectors following the Hall sensor reading configuration. In the old firmware the encoder was reset to +- 30 degrees (difference with respect to the center of the sector) respectively at each forward/backrwarrd Hall sector transition. Since the real sectors aren't equally spaced by 60 degrees, the FOC algorithm was affected by small glitches. With the new firmware the encoder is reset only on 1 to 6 and 6 to 1 transitions. Like in the sensorless configuration, the `MAXCOUNT` is reached in a complete electrical revolution. While the spacing between sectors isn't very precise, the spacing between two full electrical revolution is very precise, resulting in a very good accordance between hall sensors and encoder in tests.

PR #390, opened 2023-06-13
strain2: fix setting of orientation at startup

Nicogene:
See https://github.com/icub-tech-iit/ergocub-software/issues/129

TO BE TESTED


marcoaccame (approved):
OK @nicogene, that is OK thanks. Only thing is that your editor keeps the tab instead of converting into 4 spaces, so the code alignment is untidy. it may be worth putting the spaces instead.  I will show you how to configure the editor correctly.

PR #392, opened 2023-07-05
AMC: Increase version after fix for can discovery of strain2c

mfussi66:
As per title, the change in CAN discovery affects also the AMC board.


marcoaccame (approved):
ok, thanks

PR #395, opened 2023-07-11
Fix/battery charge

MSECode:
This PR brings the following fixes to the BAT code:
- fix the battery voltage send to the specific port to be shown in deciVolt
- align the message send from BAT to EMS board to be aligned with the BMS protocol info

### DIFF
--- a/emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
+++ b/emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
@@ -130,5 +130,6 @@ extern adc_measure_t adc_measure;
 extern adc_measure_t mean;
 extern uint32_t adc_values[9];
 extern fifo adc_samples;
+extern uint32_t vBatterydV;
 
 #endif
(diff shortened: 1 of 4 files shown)


MrAndrea (approved):
Ok

emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
valegagge, line 1: I guess the variable `vBatterydv` is used in the `main.c` file only. So the best practice is to define a local variable and not a global one.
So the `vBatterydv' could be defined in the` main.c` file inside the `case 0x5`.

This in only a suggestion, not a change request. :)

emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theBATservice.cpp
valegagge, line 1: Please remove these changes from the PR because they are meaningless and not related to your improvement.

valegagge (changes requested):
all is OK. 
For the sake of accuracy, please remove the changes in `embot_app_eth_theBATservice.cpp`

emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theBATservice.cpp
MSECode: I'll fix that

valegagge (approved):
LGTM

PR #397, opened 2023-07-17
Motor Control library used by ETH boards is now compiled in C++

marcoaccame:
The files of the motor control library used by the ETH boards are now compiled in C++ mode.

This library was originally developed to run in a C environment on the ETH boards (`ems` etc).
Over the years, some of its .c files have been added C++ model generated code and so they have been compiled in C++ mode.

Now, we need some other files to contain C++ code, so I have moved all the files into C++ compilation.
To allow a correct compilation on the ETH boards where the functions w/ prefix `MController_*` are ued by both C and C++ modules, I have done a minor change to the .h files. In particular:
- I kept the dual C and C++ linkage only in the file `Controller.h` which contains the public interface: 
  ```C++
  #ifdef __cplusplus
  extern "C" {
  #endif

  // code of `Controller.h` in here

  #ifdef __cplusplus
  }       // closing brace for extern "C"
  #endif 
  ```
- I added in file `Controller.h` an opaque struct for the `MController` which is used only via it pointer. ```
- I removed the C linkage from the other .h files (such as `Joint.h` etc) that are used only in its internal implementation;
- I added in file `Controller.h` an opaque struct for the `MController` which is used only via it pointer.

The files that use the library have been kept unchanged. Only the projects of the `ems`, `mc4plus `and `mc2plus` now compile the files `AbsEncoder.c` etc in C++ mode.

There is no change in code that requires a new version of the binaries or tests.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/examples/appl-00/proj/appl-00.uvoptx
+++ b/emBODY/eBcode/arch-arm/board/amc/examples/appl-00/proj/appl-00.uvoptx
@@ -1503,7 +1503,7 @@
 
   <Group>
     <GroupName>embot::app::eth::config</GroupName>
-    <tvExp>1</tvExp>
+    <tvExp>0</tvExp>
     <tvExpOptDlg>0</tvExpOptDlg>
     <cbSel>0</cbSel>
     <RteFlg>0</RteFlg>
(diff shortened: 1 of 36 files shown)


valegagge (approved):
LGTM

ale-git (approved):
👍

PR #398, opened 2023-07-18
Board amcbldc: fix of application version in EEPROM + TIM1 update removal optimization

marcoaccame:
This PR corrects the code of `embot::app::application::theCANagentCORE` to correctly update the application and can protocol versions into the internal storage of the board.

The change affects only the `amcbldc` board which is the only one using the above object.  I have also changed the application version to be 2.0.4.

I have tested the correct effect of the code on a dedicated setup.

Also, this PR removes the activation of the TIM1 update interrupt handler because ... it is not used and as it executes every 6 us it just wastes time.

See in here:

| amcbldc-upd-active | amcbldc-upd-deactivated |
| ------------------------------------------------------------ | ------------------------------------------------------------ |
| TIM1 UPDATE active                                           | TIM1 UPDATE disactivated                                     |

This PR is coupled w/ a similar one in here.


marcoaccame:
Thanks to @sgiraz, tests on the setup have shown that the `amcbldc` works fine, so we can merge this PR and the oen w/ the binaries.

PR #400, opened 2023-07-25
Add fix for not causing joints to go in hw fault at yri restart

MSECode:
Fix for avoiding error of hard limit reached when `yarprobotinterface` is stopped and restarted

Fault visible only on MAIS due to the restoring of default values at yarprobotinterface restart.
This is just a temporary fix, which allows to work more easily on iCub3 robots and it will be substituted by a proper patch soon. Update versioning for fw and fix error for changed method in amc


valegagge (approved):
LGTM.
Thanks!!! :)

Just a side note: we can remove the debug function `MController_Debug_Enc_values`

valegagge:
The PR with the executables file is https://github.com/robotology/icub-firmware-build/pull/95.

This PR has been tested on the iCubGenova11 successfully by @MSECode! 🚀

marcoaccame (approved):
OK, thanks

PR #401, opened 2023-07-28
2FOC: Added temperature sensor reading by I2C 

ale-git:
Overheating management with hysteresis on recover after fault.

By now the temperature limit is in the "overcurrent" parameter, in raw units, since it is not used in 2FOC.

The hysteresis is fixed to 7/8 of the limit temperature (in Celsius). 

The executables are here: https://github.com/robotology/icub-firmware-build/pull/97


marcoaccame (approved):
hi @ale-git, the change seems OK to me. I assume you have tested that well enough.

PR #399, opened 2023-07-21
Use of `embot::app::eth::theEncoderReader` in all ETH boards

marcoaccame:
## Description

This PR enables all the ETH boards to use a common interface for the the reading of the encoders used in motion control. 

It ensures:
- easy maintenance / refactoring of the code,
- easy extension towards future more powerful sensors

That is possible because we now use a PIMPL singleton object called `embot::app::eth::theEncoderReader` which is derived from an interface  `embot::app::eth::encoder::v1::IFreader` that offers functionalities as needed by today's encoders.

But it can also be derived from other interfaces to offer easy migration to encoders w/ higher resolution or to add improved diagnostics by keeping the legacy functionalities.

## Details

This PR makes the singleton `embot::app::eth::theEncoderReader` the unique point of usage for acquisition of encoders' information in motion control on ETH boards. 

It is care of the protected implementation (PIMPL) of the singleton to manage the required resources and retrieve the requested encoder values.

```mermaid
erDiagram
"motion control objects" ||..|| "embot::app::eth::theEncoderReader" : use
"embot::app::eth::theEncoderReader" ||..|| "PIMPL" : uses
"PIMPL" ||..|| "HW peripherals" : manages
"PIMPL" ||..|| "CAN located encoders" : manages
```
**Figure**. The object `embot::app::eth::theEncoderReader` and its PIMPL.

For now we use two implementations:

- One dedicated to boards such `ems`, `mc4plus` and `mc2plus` which require the `HAL` of the `stm32f407` chip. This one uses the legacy objects `EOtheEncoderReader` and `EOappEncodersReader` which can stay as they are or can be easily refactored later on.
- One dedicated to the `amc` (but later on also for the `amc2foc`) which can use the `embot::hw` framework.
### The two implementations

``` mermaid
erDiagram

"PIMPL1" ||..|| "EOtheEncoderReader" : uses
"EOtheEncoderReader" ||..|| "EOappEncodersReader" : uses
"EOappEncodersReader" ||..|| "HAL" : uses
"HAL" ||..|| "SPI encoders" : manages
"HAL" ||..|| "MOTOR encoders" : manages
"HAL" ||..|| "ADC encoders" : manages
"EOappEncodersReader" ||..|| "EOtheMAIS" : manages
"EOappEncodersReader" ||..|| "EOthePOS" : manages
"EOappEncodersReader" ||..|| "EOthePSC" : manages
```
[...]

PR #402, opened 2023-07-28
Update evaluation of final offset for calibratio14 + bugfix

MSECode:
This PR just fix a couple of bugs:
- Update evaluation on final offset to be passed by JointSet to POS service so that user set calibration5 param for calibration type 14 as it is printed when port state is read
- Add fix to bug introduced by PR #399 at [lines](https://github.com/marcoaccame/icub-firmware/blob/3a7e203eee8e40ebe333e801208b609f8eb540c4/emBODY/eBcode/arch-arm/embot/app/eth/embot_app_eth_theEncoderReaderLegacy.cpp#L82C1-L83C96)  that causes values not to be updated by state and stateExt ports

- New board version will be defined with next PR, which will brings the update for the BAT and EMS4 boards


MSECode:
First PR to be merged. Aligned with current devel version of icub-firmware-shared

MSECode:
Tested on the robot, together with the related PR on icub-firmware build.

cc: @pattacini @marcoaccame

PR #421, opened 2023-09-29
Update fw to v1.3.1 and display variable:

MSECode:
add HSM hw/fw faults flag to status_B
move sending of CAN frames from 1ms to 100ms timer if fault before next 100ms trigger sending of CAN frames in that moment if fault happens save and send to display that currents and voltages Update battery fw for sending status to display, plus triggering fault screens correctly

### DIFF
--- a/emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
+++ b/emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
@@ -13,13 +13,6 @@
 #ifndef BAT_B_H
 #define BAT_B_H
 
-// TEST VARIABLES
-extern uint32_t test_faults_final_time;
-extern uint32_t test_faults_counter_motor;
-extern uint32_t test_faults_waiting_motor;
-extern uint32_t test_faults_counter_board;
-extern uint32_t test_faults_waiting_board;
-
 extern char Firmware_vers;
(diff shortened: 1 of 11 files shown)


emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
```diff
@@ -13,6 +13,13 @@
 #ifndef BAT_B_H
 #define BAT_B_H
 
+// TEST VARIABLES
```
valegagge, line 16: Hi @MSECode,
I'm not mistaken, These variables had been used only for test purpose.
So, it would be better remove them or comment if you'll use again for other testing on banching.

valegagge (changes requested):
All is ok, since the results of the test! well done.

I suggested you a very small request for code cleaning.

emBODY/eBcode/arch-arm/board/bat/BAT_Rev_B/Inc/BAT_B.h
MSECode: thanks for the check. Good to remove it.

MSECode:
removed test variables 4ce53fcb1fc7128d3bf948889052598be7e3f20a

valegagge (approved):
Good work!!!

PR #423, opened 2023-10-05
Fix debug message

MSECode:
This PR introduces a little fix due to the missing of the info variable used in the debug messages in the method `JointSet_do_wait_calibration_6_singleJoint` in `Calibrators.c` , which lead to compilation errors in the code


valegagge (approved):
LGTM

PR #424, opened 2023-10-10
Parametrized Wrist decoupler and aea3 support in AMC board

valegagge:
In this PR the main changes are:
1. the wrist decoupler manages both the mechanical versions: MK2.0 and MK2.1.
2. support to AEA3 encoder in the AMC board.

Other details:
- Now the AMC project is official: it is in `icub-firmware-wrist\emBODY\eBcode\arch-arm\board\amc\application\v1`
- the code decoupler code is under  `WRIST_MK2 `macro, only to avoid compiling the decoupler in the EMS board. It is important to notice that the wrist decoupler can be activated by software by configuring the correct `jointset-constraint` in the configuration files
- in the AMC project, the `WRIST_MK2 `macro has been added to activate the code related to the wrist decoupler. As said before, if we want to run the controller without the wrist decoupler, we don't need to remove the `WRIST_MK2 ` macro in the project, but only set the correct parameter in the configuration files.


marcoaccame (commented):
Hi @valegagge, the changes are OK. Thanks.

The `amc` w/ the decoupler works on the setup and it is very useful to have the `amc` single project. I used it for some tests I had to do where the decoupler is not called and it works fine.

We wait till next week to complete the evaluation of execution duration of the decoupler and we shall merge afterwards.

marcoaccame:
> We wait till next week to complete the evaluation of execution duration of the decoupler and we shall merge afterwards.

Today @valegagge @sgiraz and I tested the `amc `execution time on the wrist setup and the results seems OK. 
As a result of that, we can proceed w/ the merge.

marcoaccame (approved):
OK

PR #432, opened 2023-10-26
Correct disabling of led toggling at 100ms wrongly removed in previous fw

MSECode:
Re-enabled a line for the LED toggling.


marcoaccame (approved):
Hi @MSECode, it is OK by me. If tested we can merge.

valegagge (approved):
LGTM

PR #437, opened 2023-11-06
Delete emBODY/eBdocs/arch-arm/~$D-ICUBUNIT-canprotocol-sensorboards.docx

pattacini:
Leftover.


Nicogene:
This change has been done already in `devel` and now that is merged in `master` it is not needed anymore, closing

immagine

PR #438, opened 2023-11-07
Now the joints can run based only to motor encoder - without AMO sensors - and calibrate in hard stop.

ale-git:
At present, robots like iCub 3 and ErgoCub can use AMO sensors in incremental mode with hard stop calibration.
This new version allows the same robots to run without AMO sensors at all, only relying on motor encoders.

Build PR https://github.com/robotology/icub-firmware-build/pull/112

### DIFF
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
@@ -24,27 +24,27 @@
 
 typedef struct // TripodCalib
 {
-    int32_t pwm;
-    int32_t zero;
-    int32_t max_delta;
-    int32_t start_pos[3];
+    int32_t pwm;          // [2FOC PWM units] (-32000 : +32000) = (-100% : +100%)
+    int32_t zero;         // [millimiters]
+    int32_t max_delta;    // [millimeters]
+    int32_t start_pos[3]; // [millimeters]
     
 } TripodCalib;
 
 typedef struct // CableCalib
 {
-    int32_t pwm;
-    int32_t delta;
-    int32_t target;
-    int32_t cable_range;
+    int32_t pwm;          // [MC2+ PWM units] (-3360 : +3360) = (-100% : +100%)
+    int32_t delta;        // [icubdegrees]
+    int32_t target;       // [icubdegrees]
+    int32_t cable_range;  // [icubdegrees]
 } CableCalib;
 
 typedef struct // HardStopCalib
 {
-    int32_t pwm;
-    int32_t zero;
-    int32_t space_thr;
-    int32_t time_thr;
+	  int32_t pwm;       // [2FOC PWM units] (-32000 : +32000) = (-100% : +100%)
+    int32_t zero;      // [icubdegrees]
+    int32_t space_thr; // [icubdegrees]  // placeholder if we want to make them configurable
+    int32_t time_thr;  // [milliseconds] // placeholder if we want to make them configurable
 } HardStopCalib;
 
 
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
@@ -43,8 +43,8 @@ typedef struct // HardStopCalib
 {
 	  int32_t pwm;       // [2FOC PWM units] (-32000 : +32000) = (-100% : +100%)
     int32_t zero;      // [icubdegrees]
-    int32_t space_thr; // [icubdegrees]  // placeholder if we want to make them configurable
-    int32_t time_thr;  // [milliseconds] // placeholder if we want to make them configurable
+    int32_t space_thr; // [icubdegrees]  
+    int32_t time_thr;  // [milliseconds]
 } HardStopCalib;
 
 
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
@@ -39,6 +39,14 @@ typedef struct // CableCalib
     int32_t cable_range;
 } CableCalib;
 
+typedef struct // HardStopCalib
+{
+    int32_t pwm;
+    int32_t zero;
+    int32_t space_thr;
+    int32_t time_thr;
+} HardStopCalib;
+
 
  
 #endif  // include-guard
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/Calibrators.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/Calibrators.c
@@ -596,24 +596,53 @@ BOOL JointSet_do_wait_calibration_10(JointSet* o)
     {
         int m = o->motors_of_set[k];
         
-        int e = o->encoders_of_set[k];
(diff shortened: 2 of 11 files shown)


valegagge:
Hi @ale-git,
please add a description of your changes and how they impact on the workflow of the embedded motor control. thank you!

valegagge:
https://github.com/robotology/icub-firmware/assets/4233231/8fe95e30-80e2-4bdf-bbe7-33a3fba7503d

ale-git:
In the motion control library there was still a missing step to be able to control a joint with only motor encoder, hard stop calibrated, with a EMS+2FOC architecture: before starting the hard stop calibration procedure the EMS must wait that the 2FOC itself is calibrated (rotor encoder idex found), otherwise we'll have a hard stop false detection caused by the fact that the 2FOC controlled motor doesn't move until the 2FOC is internally calibrated. In fact, the hard stop detection is base on a timeout euristics: the EMS commands a calibration PWM to the 2FOC, and if the joint doesn't move for a certain time it is assumed that the hard stop is reached. For the reason above, it was already working on MC4+ boards but not on EMS+2FOC boards. Now this PR fixes this gap, providing the right synchronization between EMS and 2FOC during hard stop calibration.

emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
```diff
@@ -1861,14 +1861,28 @@ void JointSet_calibrate(JointSet* o, uint8_t e, eOmc_calibrator_t *calibrator)
...
             
-            AbsEncoder_still_check_reset(o->absEncoder+e);
-            AbsEncoder_start_hard_stop_calibrate(o->absEncoder+e, calibrator->params.type10.calibrationZero);
-            
+            if (AbsEncoder_is_fake(o->absEncoder+e))
+            {
+                AbsEncoder_calibrate_fake(o->absEncoder+e);
+                o->hard_stop_calib.zero = calibrator->params.type10.calibrationZero;
+                o->hard_stop_calib.pwm = calibrator->params.type10.pwmlimit;
+                o->hard_stop_calib.space_thr = 120;
```
valegagge, line 1874: I assume you would have liked to use the `time_thr`here or in the next line.

emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
```diff
@@ -1861,14 +1861,28 @@ void JointSet_calibrate(JointSet* o, uint8_t e, eOmc_calibrator_t *calibrator)
...
+                o->hard_stop_calib.space_thr = 500;
+            }
+            else
+            {
+                o->hard_stop_calib.zero = calibrator->params.type10.calibrationZero;
+                o->hard_stop_calib.pwm = calibrator->params.type10.pwmlimit;
+                o->hard_stop_calib.space_thr = 12000;
```
valegagge, line 1881: same comment as the previous one

emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
```diff
@@ -39,6 +39,14 @@ typedef struct // CableCalib
     int32_t cable_range;
 } CableCalib;
 
+typedef struct // HardStopCalib
+{
+    int32_t pwm;
+    int32_t zero;
+    int32_t space_thr;
```
valegagge, line 46: Which is the measure unit? Maybe you can add a comment here nad for the field `time_thr`

emBODY/eBcode/arch-arm/embobj/plus/mc/Calibrators.c
```diff
@@ -596,24 +596,53 @@ BOOL JointSet_do_wait_calibration_10(JointSet* o)
...
+        AbsEncoder *encoder = o->absEncoder+o->encoders_of_set[k];
         
-        if (AbsEncoder_is_hard_stop_calibrating(o->absEncoder+e))
+        if (AbsEncoder_is_fake(encoder))
         {
             Motor_set_pwm_ref(o->motor+m, o->motor[m].calib_pwm);
-        
-            if (AbsEncoder_is_still(o->absEncoder+e, 12000, 1000))
+            
+            if (Motor_is_still(o->motor+m, 12000, 1000))
```
valegagge, line 605: Why here you don't use the hard_stop_calib info configured in the `JointSet_calibrate`function?

I guess the use of hardcoded values here and some lines below hides the `time_thr`configuration in the `JointSet_calibration.

valegagge (changes requested):
Hi @ale-git ,
I reported in the review comments some small changes that I think they are necessary.

After the changes, a the fw should be tested again.

valegagge:
Hi @ale-git ,
regarding the version number of `amc` fw you could find it [here](https://github.com/robotology/icub-firmware/pull/431/files#diff-d1e0d10a445c3e750355e3bc8cfc92542ff7292538bc5b569c897ced44fbee8d).

emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
ale-git: Thanks for spotting it but, as you've seen later, they are just placeholders in case that these parameters will be added to the protocol. Anyway, I don't think that they need to be made configurable since they are pretty universal and don't need to be fine tuned.

emBODY/eBcode/arch-arm/embobj/plus/mc/CalibrationHelperData.h
ale-git: ok

emBODY/eBcode/arch-arm/embobj/plus/mc/Calibrators.c
ale-git: As I said above, the parameters are just placeholders if we want make these values configurable. I'll add a comment in the code when they are defined and assigned.

ale-git:
Since the modifications consists in comments and two not used variables I'm going to rebuild the .hex (the amc has still to be done anyway), but there is no need to test it again. Generally speaking,   I suggest that our workflow should be corrected: if you want to review the PRs line by line it is well appreciated, but it should be done **before** the firmware is tested working properly, not after. Testing is a very time consuming activity, and modifications to a already tested and working version makes the previous test a useless waste of time.

valegagge:
as discussed f2f with @ale-git, we decided that in the `jointSet_calibrate` we'll initialize correctly the `hard_stop_calib` values and in the `JointSet_do_wait_calibration_10` we'll use that structure in order to avoid to leave raw values around the code. This small change should improve the maintainability and comprehension of the code.

If in the future we want to make those thresholds parametrized, we just need to set the new parameter in the `hard_stop_calib` structure in `jointSet_calibrate`.

We also agreed, that more tests are no longer necessary.

valegagge:
> Since the modifications consist in comments and two not used variables I'm going to rebuild the .hex (the amc has still to be done anyway), but there is no need to test it again. Generally speaking, I suggest that our workflow should be corrected: if you want to review the PRs line by line it is well appreciated, but it should be done **before** the firmware is tested working properly, not after. Testing is a very time consuming activity, and modifications to a already tested and working version makes the previous test a useless waste of time.

Yes, I agree with you. we need to plan testing activities better.

ale-git:
I pushed all the required changes.

valegagge (approved):
LGTM 🚀

marcoaccame (approved):
OK @ale-git, thanks

PR #443, opened 2023-12-01
Fix/battery version

MSECode:
This PR brings small change:
- Fix naming convention for BAT project so that now we have the possibility to built 2 executable:
    - a generic one called bat.hex that can work on both iCub (all versions provided with BAT) and ergoCub
    - a different version that can work only on R1 robot


valegagge:
This issue address https://github.com/robotology/icub-firmware-build/issues/115

valegagge:
Hi @MSECode,
before merging, it would be better to update the `firmware.info.xml` according to the BAT fw name change.

PR #444, opened 2023-12-06
Board `amc`: test project for streaming AEA3 values 

marcoaccame:
This PR adds some features useful for testing the reading of AEA values on the `amc` board.

## Dedicated test project that sends reading to a socket 

At first there is a dedicated test project that reads AEA3 values at a 1 ms rate and transmits them over a UDP socket to address `10.0.1.104:6666`.

The socket can be read simply by opening a shell on address  `10.0.1.1.104` and digiting
```bash
$ netcat -u 10.0.1.99 -p 6666 6666
```
and we can save the log with:
```bash
$ netcat -u 10.0.1.99 -p 6666 6666 > file.aea3.dat
```

## Using the `amc` with `yarprobotinterface`

The `amc` application can be used on the robot to move the wrist and also send AEA3 values for debug.
There are two methods.

### Log of `AEA3` values via dedicated socket
 
If the `amc` is is compiled with macro `DEBUG_AEA3_stream_over_theBackdoor` defined then it sends the AEA3 values via a dedicated socket to `10.0.1.104:6666`. Then one can get the values with `netcat` or any other reader.

### Streaming of the AEA3 values to the StateExt:0 port

If the `amc` is is compiled with macro `DEBUG_AEA3_stream_over_TORQUE` defined then it sends the AEA3 values to yarprobotinterface using the `eOmc_joint_status_t::core.measures.meas_torque` field that is not used for the wrist.
Then one can read the values directly with `yarp read` or much better with `telemetry`. See [here](https://robotology.github.io/robometry/classrobometry_1_1TelemetryDeviceDumper.html).


valegagge:
Hi,
I'm  just writing here a note about how to interpret the raw data published on the yarp port:

the raw value is transformed by embObjMotionControl [here](https://github.com/robotology/icub-main/blob/23f59fe284d3cbdf526cf2ca43d5dd6948c06547/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp#L3865), dividing the value by 10^6 as you can see from the initialization of the ` _measureConverter` object. Please see [here](https://github.com/robotology/icub-main/blob/23f59fe284d3cbdf526cf2ca43d5dd6948c06547/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp#L888C46-L888C92) and [here](https://github.com/robotology/icub-main/blob/23f59fe284d3cbdf526cf2ca43d5dd6948c06547/src/libraries/icubmod/embObjMotionControl/embObjMotionControl.cpp#L906)

I guess that the same transformation will be applied in the `TelemetryDeviceDriver` since it calls the interface function GetTorque. 

cc @marcoaccame @simeonedussoni

Issue #445, opened 2023-12-07
Make the BAT compliant with the FW update


valegagge:
Recently we noticed that when the BAT is connected to the EMS, it is not possible to discover the 2foc boards connected to the same EMS board.

We already investigated on this issue and decided to fix it making the BAT silent when the FW updater is run. In other words, the BAT starts silently on the CAN channel, while continuing to update the screen.

The EMS will send the message `SET_TXMODE` of `analog polling class` with payload 0 when the BAT service starts by YRI. 
On reception on this message, the BAT starts to send its info on CAN as its usual rate(100 ms)

When the YRI is shutting down, and all devices are closing, the EMS will send the same message but with payload 1.
On its reception, the BAT stops to send data.

NOTE: please check `TSD-ICUBUNIT-canprotocol-sensorboards.docx` for more details about the CAN message

**Dod**
- [x] EMS and BAT update
- [x] FW PR done
- [x] tests done: the BAT starts silently and it sends CAN frames only when YRI is running.
- [x] The fw update of 2foc connected to the EMS with BAT board runs successfully

MSECode:
Here the working branches of this feature:
- [icub-firmware](https://github.com/MSECode/icub-firmware/tree/feature/fwUpdaterBat)
- [icub-firmware-shared](https://github.com/MSECode/icub-firmware-shared/tree/fix/firmwareUdpater)

Basically we have done the following updates to EMS and BAT code:
- in the EMS, in the service `embot_app_eth_theBATservice` we have update the methods `can_battery_TX` and `can_battery_TXstop` in order to start and stop the CAN frame transmission of the BAT iff the ems has been started.
- in the BAT, we have filled the callback function triggered by the receiving of messages over CAN, checking that the CAN frame defining the start and stop of the transmission sent by the EMS is received and with which value. If 0 we start transmission, if 1 we stop it

In this manner we can start the BAT board as silent and we enable the transmission iff the YRI is up and running, so that we do not flood the CAN bus with messages while the ETH boards are in CAN GATEWAY MODE, while doing the Discovery in the FIrmwareUpdater

valegagge:
Hi @MSECode ,
just a clarification: your setup is composed of:
- one EMS 
-  one BAT board connected to one CAN port of EMS
- 2 2FOC boards connected to the other CAN port of EMS.

Is this correct?

In this setup, that is equal to the situation on the robot, the firmware updater and yarprobotinterface work as expected, do they?

[...]

PR #447, opened 2023-12-12
Feature/fw updater bat

MSECode:
The changes introduced with this PR are fully detailed at this link:

https://github.com/robotology/icub-firmware/issues/445

and it goes to solve this issue:

https://github.com/robotology/icub-tech-support/issues/1673

### DIFF
--- a/emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theBATservice.cpp
+++ b/emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theBATservice.cpp
@@ -474,8 +474,9 @@ eOresult_t embot::app::eth::theBATservice::Impl::Start() {
   CANmonitor::Config cfg = defaultcanmonitorconfig;
   // use ...
   cfg.target.clear();
-  // check if we are not using custom value > 100 for service.servconfig.data.as.battery.canmonitorconfig.periodofcheck 
-  // if not we are setting it to 250 in order to not having error when canmonitor checks for BAT frames
+// this if condition should be used to check if canmonitor config is passed by the configuration files
+// currently it is not woring properly since we are discussing the udpates related to this for the icub-main part
+// therefore we are keeping it commented for now and using the default configuration for the canmonitor  
 //  if(service.servconfig.data.as.battery.canmonitorconfig.periodofcheck != 0)
 //  {
 //    cfg.periodofcheck = embot::core::time1millisec * service.servconfig.data.as.battery.canmonitorconfig.periodofcheck;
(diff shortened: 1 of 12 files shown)


emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theBATservice.cpp
```diff
@@ -479,14 +474,13 @@ eOresult_t embot::app::eth::theBATservice::Impl::Start() {
...
-      embot::core::time1millisec *
-      service.servconfig.data.as.battery.canmonitorconfig.periodofcheck;
-  cfg.periodofreport =
-      embot::core::time1millisec *
-      service.servconfig.data.as.battery.canmonitorconfig.periodofreport;
-  cfg.reportmode = static_cast<CANmonitor::Report>(
-      service.servconfig.data.as.battery.canmonitorconfig.reportmode);
+  // check if we are not using custom value > 100 for service.servconfig.data.as.battery.canmonitorconfig.periodofcheck 
+  // if not we are setting it to 250 in order to not having error when canmonitor checks for BAT frames
+  if(service.servconfig.data.as.battery.canmonitorconfig.periodofcheck == 0)
```
valegagge, line 479: Hi @MSECode,
I guess that here an `else` is missing. That is, if the `periodofcheck` in the configuration sent by YRI is zero, then we need to use the default configuration, also for the other parameters. 

Moreover, the default parameters in the line 93 should be value 250 instead of 200 as you mentioned in a comment in the issue.

valegagge (changes requested):
I added a comment. What do you think?

emBODY/eBcode/arch-arm/board/ems004/appl/v2/src/eoappservices/embot_app_eth_theBATservice.cpp
MSECode: done

MSECode:
This PR is linked to:
- [icub-firmware-shared](https://github.com/robotology/icub-firmware-shared/pull/91)
- [icub-firmware-build](https://github.com/robotology/icub-firmware-build/pull/126)

valegagge (approved):
Well done! 🚀 

The EMS works properly only if the pr https://github.com/robotology/robots-configuration/pull/603 has been merged, because the battery needs the CAN address 2 as it is explained in the note `Just a more note about the need to set the address 2 to the bat board` in this [comment](https://github.com/robotology/icub-firmware/issues/445#issuecomment-1870195848)

marcoaccame (approved):
it is ok, thanks

Issue #452, opened 2023-12-19
Test the diagnsotic of the q_enc reading in 2foc boards


valegagge:
We want to validate the work done in #451 

**Dod**
- Diagnostic tested
- PR ready
- diagram about the diagnostic (sequence or state machine)

valegagge:
We want to test diagnostic simulating the lost of the index reading by sw.

valegagge:
PR : https://github.com/robotology/icub-firmware/pull/455
Diagnostic: discussed in https://github.com/robotology/icub-firmware/pull/455#issuecomment-1886862999.

Currently, we put the diagnostic info on five/wiki.

Issue #451, opened 2023-12-19
Develop the diagnostic related to the q_enc motor reading in the 2 foc


valegagge:
After validating the fix of q_enc reading bug, we want to develop a clear diagnostic that helps the user understand when e which type of error we can have on the q_enc readings.

Dod
- [x] definition of the diagnostic
- [x] implementation

valegagge:
The tests will be addressed in #452

valegagge:
Hi @ale-git, 
I know you already developed the required feature. 🚀 
Please add here the link to the branch of your code and some info related to the diagnostic.

Then you can put it in review

ale-git:
This is the draft PR.

https://github.com/robotology/icub-firmware/pull/455

ale-git:
The diagnostics is separated in two types of fault detection:
- broken index
- encoder dirty.

Notice: the "phase_broken" error type can't be detected in motors without Hall sensors, because the motor stops if the encoder stops counting. The "phase_broken" error is still detected, but only in motors equipped with quadrature encoders and Hall sensors.

This is the new diagnostics - pretty simple - moved from the interrupt service routine to the encoder reading function called at 40 kHz. 
In order to detect the dirty encoder disk it compares the expected values of the encoder (within some tolerance) and the real value when the index pin is high. In order to detect the broken index it checks the encoder value (within some tolerance) for out of range.

`
inline int QEgetPos()
{
    int poscnt = (int)POSCNT;
    
    if (QE_USE_INDEX)
    {
        static int poscnt_old = 0;

        if (qe_index_found)
        {
            if (QEICONbits.INDX)
            {
                if (poscnt_old>QE_ERR_THR && poscnt_old<QE_RESOLUTION-QE_ERR_THR)
                {
                    QE_RISE_ERROR(dirty);
                }
            }
            else if (poscnt<-16 || poscnt > QE_RESOLUTION+16)
            {
                QE_RISE_ERROR(index_broken);
            }
        }
        
        poscnt_old = poscnt;
    }

    return __builtin_divsd(((long)poscnt)<<16,QE_RESOLUTION);
}
`

https://github.com/ale-git/icub-firmware/blob/3cd6e349a5d096758819f04392b849e8e88da8b2/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/qep.c#L131

valegagge:
Dod satified. The test will be performed in #452 .

closing

Issue #446, opened 2023-12-07
Validate the q_enc reading improvement on the ergocub 1.X joints


valegagge:
# Task
During the test on ergocub 2.0 joints, we noticed that the index interrupt service routine can be executed with a big delay that is used to disrupt the FOC control at motor high speed. So we changed the position counter reset policy from software (on interrupt) to hardware (managed by the QE peripheral of the dspic33f). 

This change solved the problem on the ergocub joints 2.0.
Now we want to perform the same test on the ergocub 1.x joints

# Dod
- [x] Test done
- [x] Data collected and reported here. 
- [x] Collect the same data on ergocub 2.0 and put here the report.

For collecting data, please perform the same tests [here](https://github.com/icub-tech-iit/ergocub-design-joint/issues/179#issuecomment-1802128164).

valegagge:
Here I report the tests done on the **small joint on ergocub 2.0** highlighting the difference between running the joint with the official fw version and with the bug fixed.
![image](https://github.com/robotology/icub-firmware/assets/4233231/c07449b3-0bc6-41a1-b19b-726c89b8e259)

valegagge:
Here I report the tests done on the **torso pitch joint of ergocub 1.x** highlighting the difference between running the joint with the official fw version and with the bug fixed.

![image](https://github.com/robotology/icub-firmware/assets/4233231/16429c93-47c0-4544-bfef-9907ea2dd235)

valegagge:
cc @maggia80

maggia80:
These are amazing results!!! This means a big improvement in the performance/efficiency of our robots!

valegagge:
dod satisfied. Closing

PR #455, opened 2023-12-28
Reset quadrature encoder counter via HW

ale-git:
This PR solves a synchronisation problem in the quadrature encoder conter reset by index.
In the previous version the conter was reset by a interrupt service routine activated by the index, affected by response delay, while now the conter is reset by hardware.


valegagge:
please see [here](https://github.com/robotology/icub-firmware/issues/451#issuecomment-1880627660) for more details about the diagnostic.

valegagge:
We already did some tests using this new feature and we obtained very good results. 
https://github.com/robotology/icub-firmware/issues/446

We'll put the PR in ready after we addressed https://github.com/robotology/icub-firmware/issues/452

valegagge:
Hi @ale-git ,
just one thought: I noticed that now in the case of `encoder-dirty` or `index-broken` you call the `QE_RISE_ERROR` macro that calls the fault handler. The last sets the motor at fault. 

`#define QE_RISE_ERROR(e) gEncoderError.e=TRUE; SysError.EncoderFault=TRUE; FaultConditionsHandler()`

image



Is this correct?

If yes, in this way we are much more strict than before, when the `index broken` and the `encoder-dirty` are only warnings. 
Are we sure we want to be so severe?

ale-git:
> Hi @ale-git , just one thought: I noticed that now in the case of `encoder-dirty` or `index-broken` you call the `QE_RISE_ERROR` macro that calls the fault handler. The last sets the motor at fault.
> 
> `#define QE_RISE_ERROR(e) gEncoderError.e=TRUE; SysError.EncoderFault=TRUE; FaultConditionsHandler()`
> 
> image
> 
> Is this correct?
> 
> If yes, in this way we are much more strict than before, when the `index broken` and the `encoder-dirty` are only warnings. Are we sure we want to be so severe?

I think that if the diagnostics is reliable they should be errors and not warnings, at least the index broken. We can make the encoder dirty just a warning, but people should be forced to clean encoders when they are dirty because they spoils the performance of the control.

valegagge:
> > Hi @ale-git , just one thought: I noticed that now in the case of `encoder-dirty` or `index-broken` you call the `QE_RISE_ERROR` macro that calls the fault handler. The last sets the motor at fault.
> > `#define QE_RISE_ERROR(e) gEncoderError.e=TRUE; SysError.EncoderFault=TRUE; FaultConditionsHandler()`
> > image
> > Is this correct?
> > If yes, in this way we are much more strict than before, when the `index broken` and the `encoder-dirty` are only warnings. Are we sure we want to be so severe?
> 
> I think that if the diagnostics is reliable they should be errors and not warnings, at least the index broken. We can make the encoder dirty just a warning, but people should be forced to clean encoders when they are dirty because they spoil the performance of the control.

After a discussion with @ale-git e @MSECode,
we decided to set the `index-broken` as a warning even if we agree with @ale-git that it is an error and not a simple warning, but for now, we want to avoid the robot falling while it is walking or doing other activities.

We planned to define the `index-broken` as critical notifying the user of the problem and letting him/her find a "safe" position as soon as possible because an index broken means that the controller doesn't work properly.

ale-git:
Tested on ErgoCub arm.

`dirty` and `index_broken` encoder faults are managed as warnings. In case of `index_broken` the warning is generated and the counter is resetted by software matching it to `QE_RESOLUTION` until the synchronicity is lost. In case of `dirty` the motor continues to run compatibly with reduced performance.

Executable here https://github.com/robotology/icub-firmware-build/pull/128

valegagge (approved):
well done!!! 🚀

marcoaccame (approved):
ok, thanks

PR #459, opened 2024-01-17
Fix the control mode management in case of wrist decoupler

valegagge:
During the tests on the wrist of ergoCub, we noticed that the joints belonging to the joint set had the same control mode (HW-fault). With this PR, in case of error, each joint has its own control mode, so, for example, if an overcurrent occurs on one joint then it will be in hw-fault, while the others are in idle.

We already test this PR on the wrist setup successfully.


marcoaccame (approved):
Ok, thanks

PR #461, opened 2024-01-23
Revert the PR 459. AMC ver 1.11

valegagge:
Now the joints belonging to the same jointset go in HW if one of them is in HW fault

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v1/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v1/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {1, 10},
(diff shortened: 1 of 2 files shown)


valegagge:
I convert this PR in a draft just to test it on a wrist (robot o setup)

pattacini:
Thanks heaps @valegagge 👍🏻 

We said internally that to investigate which of the underlying motors caused the HF, one can retrieve relevant information straight from the logger, whereas the the `yarpmotorgui` will show generic messages associated with the joints.

In the future, we can enrich our diagnostics at the `yarpmotorgui` to also consider low-level motor info.

valegagge:
We tested this FW on the wrist on bench. 
We verified that if we disconnect one CAN board, all three joints go in HW-fault. If we reconnect the cable and give the idle command,  all the three joints go in position control mode.
I gonna put this pr in ready.

Thank you very much to @simeonedussoni for the tests!!!! ❤️

marcoaccame:
i gonna merge

PR #462, opened 2024-01-24
Fix namespace when calling embot partition()

mfussi66:
As per title, the compilation fails when `partition()` is called because it's in the incorrect namespace. This PR fixes it.

### DIFF
--- a/emBODY/eBcode/arch-arm/embot/app/embot_app_application_theSTRAIN.cpp
+++ b/emBODY/eBcode/arch-arm/embot/app/embot_app_application_theSTRAIN.cpp
@@ -46,7 +46,6 @@
 #include "embot_hw_pga308.h"
 #include "embot_hw_adc.h"
 #include "embot_hw_flash.h"
-
 #include "embot_hw_sys.h"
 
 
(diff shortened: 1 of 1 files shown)


Nicogene:
@davidetome do we need a new ipts release?

marcoaccame:
Hi @mfussi66, thanks for this PR. It fixes something that I missed in this former PR when I changed the scope of some FLASH functions so that we could compile simple HW test projects with a reduced number of files. 
- https://github.com/robotology/icub-firmware/pull/434

We don't need to produce new binaries.

marcoaccame:
> @davidetome do we need a new ipts release?

probably not. 

the modified file `embot_app_application_theSTRAIN.cpp` is used by the boards `strain2` and `strain2c` only to manage the FT service via CAN. The modified code manages the access of the FLASH partition used to store the calibration set of the FT sensor (aka gains, matrix, fullscales).

davidetome:
ciao @Nicogene !  As anticipated by @marcoaccame there's no need to produce a new IPTS release since the test FW is not affected by these changes 👍🏼 

cc @marcoaccame @mfussi66

PR #464, opened 2024-02-05
Use IF instead of WHILE in check motor encoder error

MSECode:
Update version to 3.3.25 and change `while` with `if` in checking of motor encoder index broken

Related to icub-firmware-build PR: https://github.com/robotology/icub-firmware-build/pull/134


valegagge (approved):
LGTM

ale-git (approved):

valegagge:
cc @marcoaccame

PR #466, opened 2024-02-14
Fix/temperature reading

MSECode:
This PR brings the following changes:
- PWM frequency reduced from 40 kHz to 20 kHz
- Improving of the management of error conditions in the temperature reading: 
    - as described in [the documentation](https://icub-tech-iit.github.io/documentation/temperature_sensors/) different negative temperatures are sent to the user to show different error conditions
    - at each error in the TDB where we are losing the correct configuration of the board we are calling its re-configuration
    - temperature sent over CAN towards the EMS in the moment the reading is done 
    - reading is now at 100 ms
    - if motor stays in overtemperature for more than 10 consecutive seconds we are putting the joint in fault, otherwise we just send warnings
    - if we cannot read from the TDB for more than 10 consecutive seconds we are putting the joint in fault, otherwise we try to reconfigure the sensor
    - the overheating error is removed when the read temperature is just less than the given threshold.
- introduced a method (never called) for generating synthetic temperature data
- now the default value for the stator-rotor offset is not 0x0000 but is 0xFFFF

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amcbldc/test/src/embot_app_skeleton_os_test.cpp
+++ b/emBODY/eBcode/arch-arm/board/amcbldc/test/src/embot_app_skeleton_os_test.cpp
@@ -1,7 +1,7 @@
 
 /*
  * Copyright (C) 2024 iCub Tech - Istituto Italiano di Tecnologia
- * Author:  Davide Tomé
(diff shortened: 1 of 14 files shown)


valegagge (approved):
Good work!
The FW has been already tested on the robot.

PR #467, opened 2024-02-14
Wrist stuck at startup bug fixed

ale-git:
The build related to this PR is here: https://github.com/robotology/icub-firmware-build/pull/138

This PR fixes the following problem: if the wrist can't reach the position set by the calibrator after the absolute encoder calibration, the joint isn't able to move any more also if it results running in position mode.

The probelm was caused by the Idle mode forced by the go-to-home failure after calibration. In fact, when put in running mode the wrist remains in go-to-safe-position, but the Idle transition has canceled its trajectory, so it can't move to its target.

Solution adopted in this PR:

The joints aren't set in `position mode` any more when they are parking, but parking makes use of its own parking trajectory generators instead.

The joint results in `calibrating` mode to the yarprobotinterface until it has completed the first parking, so that the yarprobotinterface waits before sending the position move command.

The parking mode is monitored for timeout.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v1/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v1/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {1, 11},
-            {2024, Month::Jan, Day::twentythree, 11, 00}
+            {1, 12},
+            {2024, Month::Feb, Day::twelve, 11, 00}
         },
         .OStick = 1000*embot::core::time1microsec,
         .OSstacksizeinit = 10*1024,
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
@@ -918,7 +918,17 @@ void JointSet_do_pwm_control(JointSet* o)
             }
             else
             {
-                o->control_mode = eomc_controlmode_notConfigured;
+                if ( o->wristMK2.warmup)
+                {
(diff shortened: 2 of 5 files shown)


valegagge:
I'll do the review after the tests and when it will be ready. :)

marcoaccame (approved):
Hi @ale-git, the PR is fine. And yesterday's tests on the robot are OK, right?

So, pls produce the relevant PR on icub-firmware-build and we can merge them together.

simeonedussoni:
> Hi @ale-git, the PR is fine. And yesterday's tests on the robot are OK, right?
> 
> So, pls produce the relevant PR on icub-firmware-build and we can merge them together.

@marcoaccame and all 

yesterday with @ale-git tested the wrist in startup phase. 

we got two possibilities:
1. the wrist was capable of parking smoothly and the `motorgui` was running properly
2. some of the wrist levers went in wrong direction causing the flipping of the joint: in this case the wrist could not reach the parking position

the case 2 is actually the one causing the **stuck condition** in the ergocub tests. now, with this fix, the `yarprobotinterface` streams errors stating `CAN'T EXIT SINGULARITY` and the `motorgui` is deactivated. 
It's worth noting that in some cases the wrist reaches a configuration which **resembles** the _correct_ HOME position but it isn't. This was probably the case of the failed experiments.

Thus the modification to the firmware obtains the correct behaviour: 
- if initialization/parking goes fine, than the wrist can be operated
- if initialization fails, than the GUI is clearly in error state

valegagge (approved):
Good! 🚀 
Would be great and helpful if you added comments especially related to the `warmup` field. 
Thanks

ale-git:
About the connecting rod inversion, I think that the connecting rods should be replaced with the hard stop version as soon as possible, because it is very nasty and impossible to detect by software.

PR #469, opened 2024-02-15
AMC V1.13 with improved wrist Mk2 direct kinematics solver

ale-git:
Build here: https://github.com/robotology/icub-firmware-build/pull/139

I have improved the gradient descent in the direct kinematics solver (motor angles -> yaw, pitch, roll), so that now it never reaches singularity even if brutalised with a cycle of +/- 75 degrees in pitch and roll targets, at high speed, with large overshootings.

> [!important]
The inverse kinematics algebraic solver prevents generating motor targets mechanically unreachable.

304749959-88f2a7ef-f569-4424-825d-4394c75f5352

Now the solver tries to solve for the new ypr configuration starting from the configuration of the previous control cycle, but if it leads to invalid solution it performs backtracking and reduces the step of the gradient descent. If it still fails, the initial position is resetted to (0,0,0). In this way, the average number of required steps is 4.

It is likely that all the parking procedures (startup and emergency) will result unneccessary after some time of use of the robot, and we'll be able to remove them.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v1/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v1/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {1, 12},
-            {2024, Month::Feb, Day::twelve, 11, 00}
+            {1, 13},
+            {2024, Month::Feb, Day::fifteen, 11, 00}
         },
         .OStick = 1000*embot::core::time1microsec,
         .OSstacksizeinit = 10*1024,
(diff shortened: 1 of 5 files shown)


ale-git:
The digital twin is stressed with high out-of-range angle values and overshootings, and the solver never hits singularities: 

https://github.com/robotology/icub-firmware/assets/4729270/776ba4b1-7dbe-47cf-aadf-153f37b0fbd2

ale-git:
It is likely that all the parking procedures (startup and emergency) will result unneccessary after some time of use of the robot, and we'll be able to remove them.

pattacini:
Nicely done!

> The digital twin is stressed with high out-of-range angle values and overshootings, and the solver never hits singularities:

Related PR:
- https://github.com/robotology/icub-firmware-models/pull/82#event-11811930823

marcoaccame (approved):
ok @ale-git, thanks

PR #472, opened 2024-02-21
Fix/diagnostic gen error

MSECode:
Just commented part specific to masks of generic error coming from 2FOC so that all MC generic error are then translated in icub main 

Related to:
https://github.com/robotology/icub-firmware-shared/pull/94
https://github.com/robotology/icub-main/pull/942


valegagge (approved):
LGTM

marcoaccame (approved):
ok, thanks

PR #486, opened 2024-03-15
Update amc v2.1, ems v3.86, mc4plus v3.89, mc2plus v3.68

MSECode:
Aligned the versions with the update done here: https://github.com/robotology/icub-firmware-build/pull/145


marcoaccame (approved):
great, thanks.

PR #489, opened 2024-04-08
Add jomo to diagnostic aksim. EMS ver 3.88

valegagge:
In this PR I added the value of the jomo in the par16 to make the diagnostic string more useful. 

This change affects only the EMS board because the Akism is connected to EMS on the ergoCub 2.0 test setup.

cc @MSECode


valegagge:
Hi @marcoaccame,
if you think it is better to updated the fw of the ETH board, I'll do.

I'm going to create the fw-build pr when this is ok.

Thanks

marcoaccame (approved):
ok, thanks

marcoaccame:
> Hi @marcoaccame, if you think it is better to updated the fw of the ETH board, I'll do.

as you think is best. if you need to use the binary pls do so. thanks

PR #495, opened 2024-05-23
Add comment lines for temperature reading

MSECode:
With this PR I've added some lines of comment to the file `emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/i2cTsens.c` so that one can understand what the different registries mean and how the reading and transmission of the temperatures values works in relation to the TDB board and how the transmission is managed on the I2C protocol between 2foc and TDB.

So, in this PR we just have added comments explaining the whole pipeline responsible for the temperature reading and transmission  over I2C and especially the reason why we are configuring the sensor in a specific manner.


valegagge (approved):
LGTM thanks @MSECode

PR #498, opened 2024-06-06
* This EMS 3.90 version fixes the bug that affected the stiff <-> compliant interaction mode transitions

ale-git:
* This EMS 3.90 version fixes the bug that affected the stiff <-> compliant interaction mode transitions, where the motor input type was not correctly updated in case that (for example) the stiff control output was in pwm and the torque control output was in current.

Tested on iCubGenova11 robot


ale-git:
The .hex is here https://github.com/robotology/icub-firmware-build/pull/151

marcoaccame (approved):
ok, thank you

PR #500, opened 2024-06-10
Update diagnostic for aksim2

MSECode:
This PR modifies slightly the diagnostic of the aksim2 encoder.
We are not sending the error each time an error bit is risen for the specific jomo but instead we increase a counter at each loop (which occurs at every 1ms) and then we send all the specific errors on the askim2 per each jomo one time every 10 seconds (for now - this value can be changed if needed). 
Then, on the diagnostic on the high level we print how many error per type and per jomo we got during those 10 seconds.

Diagnostic has been updated accordingly to https://github.com/robotology/icub-main/pull/968


valegagge (approved):
LGTM

marcoaccame (approved):
ok, thanks

Issue #499, opened 2024-06-07
Improve the rotor/stator calibration in 2FOC


valegagge:
After some tested carried on in https://github.com/icub-tech-iit/ergocub2-design-upperbody/issues/43, we noticed that sometimes joint couldn' t complete the calibration because it doesn't find the index before the timeout of 30 seconds exipre. 

We noticed that the motor does 3 or 4 complete tourns but the index interrupt isn't t triggered. So we need to create a setup with the oscilloscope to verify the index signal.

Moreover, in the procedure, some time constants need to be updated (or parametrized) since the 2foc now runs at 20KHz and not 40 KHz.

For more details please see [this comment](https://github.com/icub-tech-iit/ergocub2-design-upperbody/issues/43#issuecomment-2154436618).

**Dod**
The new procedure runs correctly both on the robot and on ergocub2 joints

MSECode:
After some knowledge transfer done with @ale-git, we understood that some changes should be done to the heuristic use in the method of the 2FOC where we perform the stator-rotor calibration looking for the index of the quadrature encoder mounted at the motor side of the joint.

The whole portion of code we are interested in is at this lines: https://github.com/robotology/icub-firmware/blob/master/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c#L353-L520

Here a brief explanation of what should we modified and what probably was the cause of the failing of the calibration procedure.

- First of all we need to clarify that if we do not want to find the `stator-rotor offset`, so we are just looking for the index of the encoder and setting the zero at that position, we are just passing through the steps of the state machine defined by values `0-1-2` of the variable `sAlignInProgress`. So in this case we can limit our analysis to the lines [here](https://github.com/robotology/icub-firmware/blob/master/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c#L353-L520).
- Considering this we can see a couple of variables that we need to take care of. They are: 
    - limit in the rotation of the fast shaft when looking for the index defined by [this one](https://github.com/robotology/icub-firmware/blob/master/emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c#L363)
[...]

PR #503, opened 2024-07-20
Improvement in 2FOC fw and improved quadrature_encoder diagnostic

valegagge:
**Improvements of 2foc**:
 - the I2C bus is configured to work with 100KHz speed and has some fixes in the reset procedure. More details [here](https://github.com/icub-tech-iit/study-motor-temperature/issues/20#issuecomment-2165148843). 
 - the rotor/stator calibration procedure is not more HW dependent but can run successfully on different motors. More details [here](https://github.com/robotology/icub-firmware/issues/499#issue-2340436891). 

- the information regarding the quadrature encoder diagnostic has been improved. More detail [here](https://github.com/icub-tech-iit/study-encoders/issues/23#issuecomment-2225012824)

The EMS and the other ETH boards manage the quadrature encoder diagnostic in a more user-friendly way: they count the number of errors instead of downsampling the error rate to make the final user aware of how many errors there are on a given encoder and therefore proceed with the maintenance accordingly.
This change need the change on icub-main also : PR (TBD)

cc @MSECode

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {2, 7},
-            {2024, Month::Jul, Day::four, 13, 47}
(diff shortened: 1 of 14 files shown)


marcoaccame (approved):
ok, thanks

PR #506, opened 2024-08-07
Feat/5th i2c to be merged for implementing the 5th I2C bus - first step

simeonedussoni:
in this PR there is a macro-driven implementation of the 5th I2C line. 
It's been tested on a bench setup reproducing the `ergocub` architecture, i.e. with a MTB4 boarconnected to the EMS board and updated configuration files and skinGui configuration (To Be PRed in the respective repos)

PR #509, opened 2024-09-03
Boards `mtb4` and `mtb4c`: skin on the 5-th finger finally supported

marcoaccame:
### Description of activities

The work here is a follow up of a previous [PR](https://github.com/robotology/icub-firmware/pull/506) which showed some bugs.
namely, we:
- fix of CDC offset mapping
- aligned application versions when macro `USE_FIFTH_I2C` is defined

The combined effect of these two PRs is:

- the pin connected to the fifth `I2C` channel is activated **by default**
- the project **as is** compiles and gives a **fully compatible binary file** with the previous version reading only 4 I2C channels
- to use the **fifth finger** e.g. in `ergocub` class robots, we need:
  - open the project under `....\boards\mtbX\application\proj` (X = 4 or 4c depending on the used board)
  - check the right version of comnpiler is selected (should be so) i.e. **6.19**
  - open `project options` and in the `C/C++` tab, define the `USE_FIFTH_I2C` macro
  - compile and eventually rename the binary file to distinguish it
  - at this point the binary file is ready to be flashed on the desired board, with either a programmer or the `FirmwareUpdater` tool
  - the correctness of operation can be checked from the version number which is offset by 50 wrt the _official_ one
  - the configuration files are already aligned to this on `ergocubSN001` on which the tests were done
    - the configuration may be summarized as: 
    - turn off all triangles with cdcoffset = 2200
    - turn on only triangles 0 to 4 with cdcoffset = 0000

### DIFF
--- a/emBODY/eBcode/arch-arm/board/mtb4/application/src/main-appcan.cpp
+++ b/emBODY/eBcode/arch-arm/board/mtb4/application/src/main-appcan.cpp
@@ -17,7 +17,7 @@ constexpr embot::app::theCANboardInfo::applicationInfo applInfo
 #if defined(CUSTOMIZATION_MTB4_FOR_TLR)
     embot::prot::can::versionOfAPPLICATION {20, 21, 0},
     embot::prot::can::versionOfCANPROTOCOL {20, 0}
-#elseif defined(USE_FIFTH_I2C)
+#elif defined(USE_FIFTH_I2C)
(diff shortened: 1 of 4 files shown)


simeonedussoni:
@marcoaccame I'd mark this as `ready for review` and merge, what do you think?

marcoaccame:
> @marcoaccame I'd mark this as `ready for review` and merge, what do you think?

Hi @simeonedussoni. The PR can be merged.

Issue #513, opened 2024-09-17
Improve BAT firmware for reliable battery charge for R1


MSECode:
This is a wrapper issue for https://github.com/icub-tech-iit/tickets/issues/3545

The aim is to adjust the limits used for evaluating the battery charge of R1 from the BAT.
A side note to be done is that the evaluated battery charge is not completely reliable since it is based on just the mean voltage and it is linear between its max and min.

The lower and upper bound have been defined as already done for ergoCub and as shown at this comment: https://github.com/robotology/icub-firmware/issues/476#issuecomment-1980985493

I've used the same starting values (which are then differently weighted by the number of cells we have in series) since the cell used is the same, i.e. Panasonic NCR18650

MSECode:
As requested by the ticket listed and following the decision previously made and described at the linked comment, we can open the PR on firmware and firmware-build. Basically, in order to provide a more reliable evaluation of the battery charge for R1 coming from the BAT we adjust the lower threshold.

PR are linked here:

- icub-firmware: https://github.com/robotology/icub-firmware/pull/514
- icub-firmware-build: https://github.com/robotology/icub-firmware-build/pull/162


Let me know if the changes are fine, so I can move the PRs in ready. I'll keep them in draft for now even if the changes is minimal.

cc: @maggia80 @valegagge

PR #514, opened 2024-09-17
Adjust lower bound for evaluating battery charge for R1

MSECode:
As described at this issue: https://github.com/robotology/icub-firmware/issues/513 we have chosen to adjust the lower bound for evaluating the battery charge in R1.
The estimation, however, cannot be consider as perfect since it is based just on the voltage reaching the BAT board and moreover it is linear and not based on a proper coulomb counter, as done by a BMS.

cc: @valegagge @maggia80


maggia80 (approved):

valegagge (approved):
LGTM

valegagge:
@MSECode please also prepare the PR on icub-fw-build.
thanks heap!

valegagge:
this is the related PR in icub-fw-build: https://github.com/robotology/icub-firmware-build/pull/162

PR #525, opened 2024-10-16
MTB4/c implementation of 5th I2C line can be switched runtime with CAN message

simeonedussoni:
main feats introduced

- a new skintype is added with `type = 3` to deal with switching of triangle reshuffle
- all the part dealing with the compensation is updated. 
- the compensation also takes care of the two different types of boards used in ergoCub hand i.e. 
  - the fingertips, which need no temperature drift correction
  - the palm, which may need it.
sending a CAN message `4D` with firts byte equal to 3 enables the 5th I2C remappijng to peruse the triangles with Id >15 and include them in the CAN stream

tested with CANREAL and ESD CAN tool. 

TBD: implementation of CAN mesg forming in the EMS

### DIFF
--- a/emBODY/eBcode/arch-arm/board/mtb4/application/proj/mtb4-application-v6.uvoptx
+++ b/emBODY/eBcode/arch-arm/board/mtb4/application/proj/mtb4-application-v6.uvoptx
@@ -754,7 +754,7 @@
 
   <Group>
     <GroupName>embot-core</GroupName>
-    <tvExp>1</tvExp>
+    <tvExp>0</tvExp>
     <tvExpOptDlg>0</tvExpOptDlg>
     <cbSel>0</cbSel>
     <RteFlg>0</RteFlg>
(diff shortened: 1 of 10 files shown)


marcoaccame (changes requested):
Hi @simeonedussoni I see too many changed files that are not relevant to the `mtb4`: in folders `amc2c`, `amcbldc`, `mbd`. 

They must be removed from the PR. Maybe you just need to rebase your branch vs the latest `robotology/devel`.

simeonedussoni:
hii @marcoaccame 
maybe you compared wioth `master,` I see only 2 commits w/ 10 files modified, all related to the MTB4/C board.
This is consistent with the fact I deleted and recreated my fork just before doing this activity to insure a cleaner history, but actually if I look at the branch status it comapres to `master` rather than `devel`  

Actually I see that there is also the project files `.uvproj` changed but don't need that. maybe it's worth restoring the `devel` versions of these files

PR #538, opened 2024-11-27
Fix of BSP of the board `amcfoc` to run EEPROM

SanLordKevin:
This PR fixes the SPI 3 pin configuration inside `embot::hw::spi::bsp`. Now the `embot::hw::eeprom` works fine.

I have also  inserted some macros in `main.cpp` of the test project to better test the code.

There are some macros and portion of the code commented for testing, I left them commented as their use depend on the specific test to do.


marcoaccame (approved):
It is fine. 

We can merge as these changes touch only the `amcfoc` board.

PR #545, opened 2024-12-05
`amcfoc` test application

SanLordKevin:
Adding test application project for the `amcfoc` on `cm4` core.
The starting projects were taken from the `amc` application folder and from the UDP demo for `amcfoc`.


marcoaccame (approved):
The PR is OK. I only cleaned up a bit.

It can be merged as it touches only files used by the `amcfoc` that is WIP.

PR #546, opened 2024-12-10
`amcfoc` info in loader

SanLordKevin:
Inserted the right info for the `amcfoc` in the `loader` project.

As a sidenote i removed a breakpoint in the project.


marcoaccame (approved):
that's ok, thanks

PR #548, opened 2024-12-12
`amcfoc` timer BSP implementation

SanLordKevin:
Corrected the timer BSP for the `amcfoc`, also added the code and macros for testing the timer,


marcoaccame (approved):
ok, thanks

PR #550, opened 2024-12-19
Board `amcfoc`: draft `app.yri` and refinement of `app.mot`

marcoaccame:
The `amcfoc` board has now a draft `app.yri` and a refined  `app.mot`.

The former is the one that talks to `YRI` and then communicates to the second either via ICC or via CAN.

The latter correctly execute motor control of the `speedgoat` emulated motor in voltage control mode and current control mode.

This PR can be safely merged because does not touch any other code / service of boards currently used in teh robot


emBODY/eBcode/arch-arm/embot/app/bldc/embot_app_bldc_theCOMMcanicc.cpp
```diff
@@ -292,7 +292,8 @@ void embot::app::bldc::theCOMM::Impl::tCOMM_OnEvent(embot::os::Thread *t, embot:
             remainingINrx = 0;   
             if(true == embot::app::eth::theICCservice::getInstance().get(item, remainingINrx))
             {
-                impl->tCOMM_OnRXmessage(t, eventmask, param, {item.des, item.frame}, impl->_tCOMMoutmessages); 
+                embot::app::eth::theICCservice::ItemCANframe icf {item};
```
marcoaccame, line 295: memo: something strange. I should not have a `embot::app::eth::theICCservice::ItemCANframe` in here. 
i must review why i see it.

marcoaccame: OK, I solved. I committed an experimental change in ICC that works but I prefer to test for some more time. So, I will revert it.

But let me explain what ii is about.

The `embot::app::eth::theICCservice` now transports an `Item` object that is just a CAN frame:
```C++
        struct Item
        {
            embot::app::msg::Location des {};
            embot::prot::can::Frame frame {};
            constexpr Item() = default;
        };
```
It would be much better for it to transport a generic `Item` object able to contain some bytes and then to fill in it a CAN frame object or any other we need, such as in the following.

```C++
        enum class ItemType : uint8_t { RAW = 0, CANframe = 1, ROP = 2 };
          
        struct Item
        {
            ItemType type {ItemType::RAW};
            uint8_t filler[3] = {0};
            uint8_t data[20] = {0};
            constexpr Item() = default;
        }; 
        static_assert(24 == sizeof(Item), "sizeof(Item) is not 24");      

        struct ItemCANframe : public Item
        {
           ...
        };

        struct ItemROP : public Item
        {
            enum class CMD : uint8_t { ask = 0, say = 1, set = 2, sig = 3, ack = 4, nak = 5 };
            ....
        };
```

With the above we can offer current functionality of transport of CAN frames for MC but also other needs such as: ping the other application on the other core, ask its version, ask data available only to the other core such as the UID needed by the MAC address.

But that will not be a matter of the PR.

PR #551, opened 2025-01-08
2foc.3.3.38: solves the overcurrent problem

ale-git:
It solves the overcurrent problem by disabling the current sensor gain calibration at startup.


marcoaccame (approved):
ok, thanks

PR #552, opened 2025-01-08
Fix/hand skin fix: correct I2C mapping for best cablilng

simeonedussoni:
as per F2F interactions with the cabling guys, I slightly modified the I2C mapping for the `skinType  = 3` case.

this involves both MTB4 and MTB4c thus I also incremented the minor version fo both boards. 

related PR issued on `icub-firmware-build` repo to keep alignment


simeonedussoni:
motivations for the PR reported [here](https://github.com/robotology/icub-firmware-build/pull/188)

marcoaccame (approved):
OK @simeonedussoni, thanks

PR #553, opened 2025-01-09
bugfix `amcfoc` app.yri and eupdater

SanLordKevin:
Fix the position of the application information in the memory of `amcfoc mot app`. 
Resolve a bug in the `eupdater` information.


marcoaccame (approved):
ok, thanks

PR #557, opened 2025-01-16
Rebuild of amc, mc2plus, mc4plus after introduction of LuGre 

Nicogene:
This was a regression due to: https://github.com/robotology/icub-firmware-shared/pull/99


marcoaccame (approved):
ok, thanks

PR #564, opened 2025-02-04
Work on update of AMO encoder diagnostic

MSECode:
This PR updates AMO encoder diagnostic so that it might be easier for a user to identify the error and fix errors for overheating and i2c comm failure even if now we still send the old error message even when we loose i2c communication on the TDB and update fw versions

### DIFF
--- a/emBODY/eBcode/arch-arm/embobj/plus/board/EOappEncodersReader.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/board/EOappEncodersReader.c
@@ -42,6 +42,7 @@
 #include "hal_spiencoder.h"
 #include "hal_quadencoder.h"
 #include "hal_adc.h"
+#include "embot_core_binary.h"
 
 #include "EOtheEntities.h"
 #include "EOconstarray.h"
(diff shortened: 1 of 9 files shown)


valegagge:
I'll check this pr in the next few days.

MSECode:
> I'll check this pr in the next few days.

No problem since I need first to wait for https://github.com/robotology/icub-firmware-build/pull/193 to be closed.

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
```diff
@@ -959,6 +965,15 @@ static void Motor_raise_fault_overheating(Motor* o)
     o->control_mode = icubCanProto_controlmode_hwFault;
 }
 
+static void Motor_raise_fault_i2c_communication_failure(Motor* o)
```
valegagge, line 968: Is this function useful? If I'm not wrong this function isn't called anywhere.  

Note the same for `Motor_raise_fault_overheating`.

valegagge (changes requested):
Hi @MSECode, I think here the fill of error param in `s_eo_appEncReader_IsValidValue_AMO` is missing. Please could you check? thanks

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
MSECode: That's true. Looking at the code I should remove also `static void Motor_raise_fault_overheating(Motor* o)` some lines above. I've realized that actually that error, being managed by the 2foc, is just checked by the method `MController_update_motor_state_fbk(jointindex, frame->data);` where we receive the whole error mask sent by the 2FOC via CAN and check which bits have been set to 1. Thus we can remove that static method.

emBODY/eBcode/arch-arm/embobj/plus/board/EOappEncodersReader.c
```diff
@@ -1497,6 +1565,67 @@ static eObool_t s_eo_appEncReader_IsValidValue_AKSIM2(uint8_t jomo, hal_spiencod
     return ret;
 }
 
+static eObool_t s_eo_appEncReader_IsValidValue_AMO(uint8_t jomo, hal_spiencoder_diagnostic_t* diag, eOencoderreader_errortype_t *error)
+{
```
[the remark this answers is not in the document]
MSECode: In this case, since we are using the same enum type for all errors I set the error in case that method returns false. I left the signature equal to the one used for the other encoders in case in future we wanna update the enum with other error values that should then be updated inside the method call

emBODY/eBcode/arch-arm/embobj/plus/mc/Motor.c
valegagge: Functions removed. thanks @MSECode

valegagge (approved):
LGTM

PR #565, opened 2025-02-21
Fix error for Aksim2 and AMO encoders

MSECode:
This PR fixes a problem with some missing error cases used in the switch loop that might generate o->fault_state.bits.data_error in fw. Those error cases were not added to the switch since we are managing those differently using diagnostic functions in other section of the code. However, the default case of the switch is returning o->fault_state.bits.data_error, thus this fix was mandatory.

Specifically we are providing a solution to the bug in `AbsEncoder_invalid` where a `swtich-case` loop set different error bits depending on the error passed. However, since the default case sets the fault bit `o->fault_state.bits.data_error`, all the non managed errors make that bit to be set to 1. 
Thus Aksim2 and AMO errors, that are managed by a particular diagnostic, made that bit to be up. Those we need to managed those errors values also here.

Moreover, we are fixing the bug related to this issue: https://github.com/icub-tech-iit/ergocub2-design-upperbody/issues/42 that was caused by non filtering spikes in joint position values when using an Aksim2 encoder.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {2, 17},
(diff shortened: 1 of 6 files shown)


valegagge (approved):
LGTM

PR #566, opened 2025-02-26
`amc` and` amcfoc` encoder BSP refactoring

SanLordKevin:
`amc` and` amcfoc` encoder BSP refactoring

added possibility to custom SPI configuration depending on the board

`amc` SPI BUS clock modified, now is 100MHz

MA730, aea3, now reads 2 bytes instead of 3, bug correction


emBODY/eBcode/arch-arm/embot/hw/embot_hw_chip_MA730.cpp
```diff
@@ -111,7 +111,7 @@ struct embot::hw::chip::MA730::Impl
 {                
     bool _initted {false};
     Config _config {};
-    uint8_t _databuffer[3] = {0, 0, 0};
```
marcoaccame, line 114: OK, great. I want to stress that you found it out by using the oscilloscope that showed a SPI transaction of 3 bytes rather than the expected two.

emBODY/eBcode/arch-arm/embot/hw/embot_hw_encoder_bsp.h
```diff
@@ -27,9 +27,10 @@ namespace embot::hw::encoder {
     struct PROP
     {   // this one is ok for a spi encoder.
         embot::hw::SPI spiBus {embot::hw::SPI::none};
+        std::array<embot::hw::spi::Config, embot::hw::encoder::numberofSPIencoders> spicfg {};
```
marcoaccame, line 30: OK, and in here is the configuration of the SPI for the given encoder

marcoaccame (approved):
ok, thanks @SanLordKevin. The changes are OK. I have edited however the title and the basic description so that it can be clear what has been done.

Issue #554, opened 2025-01-13
Current set to zero after switching the control mode to current


GiulioRomualdi:
I’m currently working on the current control mode and noticed an issue. When switching the control mode from position to current, the setpoint is set to 0.

However, when switching from position to torque, the setpoint in torque is set to match the estimated value.

After discussing this with @valegagge, it seems this might be a bug in the 2foc implementation, where the setpoint in the current mode is not being set correctly.

Here, it's a video showing the behaviour 

https://github.com/user-attachments/assets/c1a9356a-9321-414f-91fd-5f50b53d91e2

GiulioRomualdi:
To provide a high-level perspective: The problem with this behavior is that when we switch to current mode while the robot is on the ground, the desired current is momentarily set to zero. Although this happens for a very short time, it is sufficient for the robot to collapse under its own weight.

GiulioRomualdi:
Cc @isorrentino @giulero @loremoretti

Nicogene:
Ccing @ale-git that maybe knows what is happening

valegagge:
I provided the ems fw here attached to @GiulioRomualdi to test on the robot. 
Currently, the fw contains only a debug message to understand in detail what happening in the current control mode switching. 

![Image](https://github.com/user-attachments/assets/a63cb9ca-e3e7-4838-9edf-c2cc60687d60)


Please note the the fw I provided is based on the current version on the robot (3.95) and not on the top of the devel branch.

Thanks @GiulioRomualdi for the tests!

[ems_x_giulio_v_3_195.zip](https://github.com/user-attachments/files/18441345/ems_x_giulio_v_3_195.zip)

GiulioRomualdi:
I'm currently testing the firmware on ergoCubSN002. I flashed the board 10.0.1.6 the right upper leg

![Image](https://github.com/user-attachments/assets/44522b19-9151-4793-8ce7-1a15826bf132)


Here the output of the interface
[interface_current.txt](https://github.com/user-attachments/files/18441908/interface_current.txt)

This is the video of the experiment

https://github.com/user-attachments/assets/94967be4-3f65-4e5d-a271-a7927c523d17

valegagge:
Hi @GiulioRomualdi ,
is it clear from your test that the references provided to the pid are 0 as we supposed before.

Now the solution is to provide the current value of `current`, but... we need to filter it in some way, because its values change a lot!

GiulioRomualdi:
[...]

PR #569, opened 2025-02-27
On switch current control, the target current has the value of the current before the switch

valegagge:
This PR has the aim to fix the behavior explained here: https://github.com/robotology/icub-firmware/issues/554.

In a few words, now on the current control switch, the embedded control mode sets as target the value of the current before the switching instead of zero.

This new behavior helps when the robot is standing up and the walking-control application switches from position to current by preventing the robot falls.

This fw has been tested on the robot and setup.


valegagge:
as soon as the PR is ok I'll provide the icub-fw-build PRs

marcoaccame (approved):
Hi @valegagge, the PR is OK. thanks

PR #570, opened 2025-03-05
Update amcbld test project

MSECode:
This PR adds the following updates:
- Add tests for option bytes and for PWM1, PWM2, PWM3 
- Update porject file adding correct stm32hal.g4.v122 lib file and removing *.v120
- Update methods for reading encoder and some hw values that now use embot wrapper function istead of calling directly HAL methods


valegagge (approved):
LGTM

thanks

PR #575, opened 2025-03-10
`amcfoc` analog readings

SanLordKevin:
This PR adds the reading of the `VIN` of the board, so that the `amcfoc.mot` application can use this value and pass it to the MBD motor control code.
The `VIN` value is retrieved using the `embot::hw::motor::getVIN()` function that uses at lower level the `ADC3` triggered by the `TIM15` timer programmed at 10 ms tick.

The `ADC3` reads also other quantities other than the `VIN`:
```c++
volatile float       AinInputCurrent;           /* Input current (A) */
volatile float       AinInputVoltage;           /* Input voltage (V) */
volatile float       AinVaux;                   /* Auxiliary +5V voltage (V) */
volatile float       AinVcc;                    /* Main 3.3V voltage (V) */
volatile float       AinVcore;                  /* Core 1.2V voltage (V) */
volatile float       AinCoreTemp;               /* Core temperature (C) */
volatile float       AinMot1Temp;               /* Motor 1 temperature (C) */
volatile float       AinMot2Temp;               /* Motor 2 temperature (C) */
```
Note: `AinCoreTemp` reading is not implemented yet. `AinMot1Temp` and `AinMot2Temp` reading has not been tested.

The code had been tested on a setup with different voltage power supply, the difference between the value set and the reading was around 1%.


marcoaccame (approved):
ok, thanks

PR #579, opened 2025-03-18
dspic33f non standard short summation fixed (current filter bug).

ale-git:
The 2FOC dspic33f processor doesn't respect the C standard for integer promotion in summation with short type: instead of promoting the result to long and truncating it if assigned to a short type, it truncates the result in any case. Fixed and tested on SN 0001. Added current feedback reset when the motor drive is disabled.


marcoaccame (approved):
ok, @ale-git

ale-git:
@traversaro yes, the last good release was 3.3.36.

PR #580, opened 2025-03-20
Add i2c communication error condition handling

MSECode:
This PR introduces the fix of rising the correct error bit of the `SystemError` bitmask when the 2foc i2c bus cannot read any temperature values coming from the tdb board for more than 10 consecutive seconds. This basically fix the previous way of handling the error where we were using the same error message for managing 2 different conditions: this one and the the actual overheating error, i.e. when the temperature read is over a threshold set.
This update of the 2foc fw aligns with the the updates already developed and merged made on the ems fw and on the sw side in `icub-main` available here: https://github.com/robotology/icub-firmware-build/pull/192


ale-git (approved):

marcoaccame (approved):
ok @MSECode that is fine.
I know that you have just tested that again today so I will merge this and the associated PR:
 - https://github.com/robotology/icub-firmware-build/pull/202

MSECode:
Exactly, as shown in the screen below the error is correctly shown on both `yri` and `motorgui`. I've triggered it by disconnecting the TDB cable

image
 
Now the user can correctly be informed if a real overheating is happening or the tdb problem is just related to a i2c bus communication issue.

PR #581, opened 2025-03-25
Add project for test on amc board

MSECode:
This PR starts the work that need to be carried out on the amc board, where we are gonna implement the tests that are going to be implemented on both CM4 and CM7 cores. For this reason we are going to define 2 projects:

- the first one is based on the amc template where we are enabling only the cm7 routine
- the second is based on the amc2c template, thus it will enable both cm7 and cm4 routines

In order to make the projects that should be robust, maintainable and that can exploit structures that are common between projects (avoiding duplications) we organized the classes as described in brief in here: https://github.com/icub-tech-iit/ipts/issues/57

As things now, this PR actually add to the main code the 2 project and implements some of the tests for the amc board that have been detailed in short in this issue: https://github.com/icub-tech-iit/ipts/issues/56.


marcoaccame (approved):
ok, thanks

PR #582, opened 2025-03-27
Fix error management for aksim2 joint

MSECode:
This PR solves a problem we had with Aksim2 encoders due to the fact that their diagnostic is managed differently from all other encoders and the error risen need to be filtered accordingly without making the AbsEncoder class managing them.
This problem generated only on our bench setup, which uses joints that mount an `Aksim2` encoder.
Briefly what was happening is that if the number of errors for the aksim was over a specific threshold we were always getting an hw fault at the joint, without knowing the reason due to the fact that now error message was transmitted. 
The source of the error is related to these lines of code: https://github.com/robotology/icub-firmware/blob/master/emBODY/eBcode/arch-arm/embobj/plus/mc/AbsEncoder.c#L418-L464
as in detail to this id condition: https://github.com/robotology/icub-firmware/blob/master/emBODY/eBcode/arch-arm/embobj/plus/mc/AbsEncoder.c#L428-L431.
Thus due to the facts that:
1. With Aksim2 encoder we manage the errors and their diagnostic is a specific way because of the high number of crc error
2. A generic error is sent when a certain general threshold is overcome

we were having the joint in fault without knowing the specific reason while reading from  `Aksim2` encoders. 

Related to: https://github.com/robotology/icub-firmware-build/pull/204
Update ems v3.103, mc4plus v3.103, mc2plus v3.82, amc v2.20, amc2c v3.5.0

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {2, 19},
-            {2025, Month::Mar, Day::seven, 8, 00}
+            {2, 20},
+            {2025, Month::Mar, Day::twentyfive, 17, 17}
         },
         .OStick = 1000*embot::core::time1microsec,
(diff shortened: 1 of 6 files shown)


marcoaccame (approved):
ok, thanks

PR #584, opened 2025-04-01
2FOC PWM_MAX increased to 96%.

ale-git:
When the maximum pwm was increased to 96%, there was still another check at lower level. Fixed.

Version increased to 3.3.42


marcoaccame (approved):
OK, thanks. I am aware it was tested in ergoCub SN000

PR #585, opened 2025-04-04
CRC errors detection in `amc` board: support

SanLordKevin:
This PR adds support for CRC error detection in `amc` board.

The CRC error status is read through the SPI bus from the ETH switch and if any errors are found, then `theETHmonitor` sends diagnostics messages to `yarprobotinterface`.

So far, test were done using a fake generation of CRC errors and  `yarprobotinterface` correctly displays the diagnostics as follows.

image

The effective reading from the register must be verified in presence of CRC errors induced in the ETH network. That will be done a later moment.

This change does not affect normal behavior, so it can be safely merged.


marcoaccame (approved):
that is fine, thanks

Issue #576, opened 2025-03-10
Increase safety checks for joint Torque/Current control to prevent mechanical limit breakage


isorrentino:
Currently, when joints are controlled in torque or current mode, there are insufficient safety checks to prevent them from reaching or exceeding mechanical limits. This has already led to multiple instances of joints breaking their mechanical limits due to the absence of proper constraints.

List of recent issues reporting this problem:
- https://github.com/robotology/icub-tech-support/issues/2054
- https://github.com/robotology/icub-tech-support/issues/2049
- https://github.com/robotology/icub-tech-support/issues/2004

cc @maggia80 @GiulioRomualdi @traversaro @DanielePucci @Fabrizio69

ale-git:
Hi @isorrentino,
when the control mode is "torque", or the interaction is "compliant", the firmware prevents to reach the hardware limit by implicitly and temporarily switching to "position" mode when the joint exceeds the software position limit.
About the "low level" control modes pwm and current, this part of the firmware is very old, when things where decided F2F and usually not documented, so I have to rely on my memory. The request was that the firmware acted as a simple bidirectional pipeline for motor commands and sensor readings, transferring all the responsibility of intelligent management of the joint to the software client. I'm not particularly fan of this choice, I'm just explaining why there is no position limit control in these modes. Anyway, it should be quite easy to extend to "pwm" and "current" modes the same check performed with torque/compliant mode/interaction.

isorrentino:
> Anyway, it should be quite easy to extend to "pwm" and "current" modes the same check performed with torque/compliant mode/interaction.

Thanks a lot, @ale-git, for your clarification. I noticed that when the robot is controlled in current mode, it always pushes against its limits, but now I understand why. I'm not sure about torque control mode, as I usually use it at a higher level and then send current references. I'll add additional checks at the high level, but it would also be beneficial to prevent these issues at the low level.

isorrentino:
It happened again https://github.com/robotology/icub-tech-support/issues/2058.

We noticed that if the high-level application stops for any reason without switching the control mode, the low-level continues to apply the last desired reference current (which is how the hardware limit was broken yesterday). This should actually be handled differently.

cc @Nicogene

pattacini:
Hi all,

While trying to set the stage for tomorrow's meetup, I came across these two papers from AMI's legacy:
- "[Torque Control with Joints Position and Velocity Limits Avoidance](https://ieeexplore.ieee.org/document/10160693)"[^1]
[...]

PR #586, opened 2025-04-04
Added setpoint reference timeouts to PWM, current and torque.

ale-git:
If a watchdog expires, the joint is put in stiff position mode. The torque feedback timeout behaviour changes from go to hardware fault to switch to stiff position mode.

### DIFF
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
@@ -609,6 +609,12 @@ extern void Joint_clear_faults(Joint* o)
 {
     o->fault_state_prec.bitmask = 0;
     o->fault_state.bitmask = 0;
+    
+    WatchDog_rearm(&o->vel_ref_wdog);
+    WatchDog_rearm(&o->trq_ref_wdog);
+    WatchDog_rearm(&o->pwm_ref_wdog);
+    WatchDog_rearm(&o->cur_ref_wdog);
+    WatchDog_rearm(&o->trq_fbk_wdog);
 }
 
 int8_t Joint_check_limits(Joint* o)
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
@@ -510,13 +510,13 @@ BOOL Joint_check_faults(Joint* o)
     if (o->fault_state.bitmask != o->fault_state_prec.bitmask)
     {        
         static eOerrmanDescriptor_t descriptor = {0};
+        descriptor.par16 = o->ID;
         descriptor.par64 = 0;
         descriptor.sourcedevice = eo_errman_sourcedevice_localboard;
         descriptor.sourceaddress = 0;
                     
         if (o->fault_state.bits.torque_sensor_timeout && !o->fault_state_prec.bits.torque_sensor_timeout)
         {   
-            descriptor.par16 = o->ID;
             descriptor.code = eoerror_code_get(eoerror_category_MotionControl, eoerror_value_MC_axis_torque_sens);
             eo_errman_Error(eo_errman_GetHandle(), eo_errortype_error, NULL, NULL, &descriptor);
             
@@ -530,7 +530,7 @@ BOOL Joint_check_faults(Joint* o)
         
         if (o->fault_state.bits.torque_ref_timeout && !o->fault_state_prec.bits.torque_ref_timeout)
         {   
-            descriptor.par16 = eoerror_value_MC_ref_timeout_torque;
+            descriptor.par64 = eoerror_value_MC_ref_timeout_torque;
             descriptor.code = eoerror_code_get(eoerror_category_MotionControl, eoerror_value_MC_ref_setpoint_timeout);
             eo_errman_Error(eo_errman_GetHandle(), eo_errortype_error, NULL, NULL, &descriptor);
             
@@ -544,7 +544,7 @@ BOOL Joint_check_faults(Joint* o)
         
         if (o->fault_state.bits.current_ref_timeout && !o->fault_state_prec.bits.current_ref_timeout)
         {   
-            descriptor.par16 = eoerror_value_MC_ref_timeout_current;
+            descriptor.par64 = eoerror_value_MC_ref_timeout_current;
             descriptor.code = eoerror_code_get(eoerror_category_MotionControl, eoerror_value_MC_ref_setpoint_timeout);
             eo_errman_Error(eo_errman_GetHandle(), eo_errortype_error, NULL, NULL, &descriptor);
             
@@ -558,7 +558,7 @@ BOOL Joint_check_faults(Joint* o)
         
         if (o->fault_state.bits.pwm_ref_timeout && !o->fault_state_prec.bits.pwm_ref_timeout)
         {   
-            descriptor.par16 = eoerror_value_MC_ref_timeout_pwm;
+            descriptor.par64 = eoerror_value_MC_ref_timeout_pwm;
             descriptor.code = eoerror_code_get(eoerror_category_MotionControl, eoerror_value_MC_ref_setpoint_timeout);
             eo_errman_Error(eo_errman_GetHandle(), eo_errortype_error, NULL, NULL, &descriptor);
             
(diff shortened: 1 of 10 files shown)


Nicogene:
Hi @valegagge,

The issue reference is this one:
- https://github.com/robotology/icub-firmware/issues/576

Here instead you can find some action points about this activity:
- https://github.com/robotology/icub-firmware/issues/576#issuecomment-2737055518

One of them was to add a watchdog also on PWM, current and torque control in order to avoid that in case of fault it keeps the setpoint and lead to a possible breakage

valegagge:
Tomorrow @ale-git will test on the robot. I put this PR in draft

Nicogene:
Hi @ale-git, @valegagge!

How did the test on the robot go? Can we merge this PR after resolving the conflicts?

valegagge:
I didn't know if @ale-git have chance to test it on the robot today. He was working on higher priority task.

ale-git:
I'm going to test it on monday, Jacopo reserved a robot.

ale-git:
I've tested the automatic switch to `position/stiff` when:

- a watchdog timeout occurs for setpoints in `openloop` and `current` control modes
- a watchdog timeout occurs for feedback in `torque` control mode

Moreover, when the joint is in `velocity` control mode the motion is stopped by setting `velocity = 0`, but this behaviour was already implemented. Also the watchdog of the feedback in `torque` control mode was already implemented, but the behavior has changed, i.e. in this new version we switch to `position/stiff` instead of a `fault` state. 

The test yarp module changes the control mode of the ankle joints every 10 seconds, but sends setpoints (equal to 0) only for 5 seconds. Thus we see in the video below a cycle of `pwm/current/velocity/torque` modes, each one switching to `position/stiff` after 5 seconds and staying there for other 5 seconds - except for velocity mode, that doesn't have to switch to `position/stiff`.

https://github.com/user-attachments/assets/6f5ba4c4-24be-4cbc-ad61-ccb937e1979a

emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
```diff
@@ -391,22 +412,51 @@ void Joint_update_torque_fbk(Joint* o, CTRL_UNITS trq_fbk)
...
+            *errorcode = eoerror_code_get(eoerror_category_MotionControl, eoerror_value_MC_axis_torque_sens);             
+            return TRUE;
+        }
+    }
+
+    if (o->control_mode == eomc_controlmode_torque)
+    {
+        if (WatchDog_check_expired(&o->trq_ref_wdog))
+        {  
+            *errorcode = eoerror_code_get(eoerror_category_MotionControl, eoerror_value_MC_trq_ref_timeout);
```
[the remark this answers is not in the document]
MSECode: I can resolve this conversation. I did not recall that the update was done in here: https://github.com/robotology/icub-firmware-shared/pull/110

ale-git:
> 1. What happens if the joint is moving in current or PWM and the timeout expires? does the fw update the target position accordingly? Did you test it?

Yes, sure. When the joint is set in `position` controlmode, the joint position target is automatically set to the current position as usual, the same thing that happens when you switch to `position` from any other control mode by the yarpmotorgui.

https://github.com/robotology/icub-firmware/blob/550da1424cd9f049665782e7c6db4847e69f3203/emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c#L358

> 2\. How did you measure the [maximum latency](https://github.com/robotology/icub-firmware/pull/586#issuecomment-2909555846)?

I've added a variable inside the watchdog struct to keep track of the minimum value reached by the timer, and it is updated automatically every time the watchdog is rearmed. Thus, the value calculated is the maximum latency over the whole experiment.

ale-git:
In order to avoid the yarpmotorgui incompatibility, I've added the setpoint timeout configuration to the protocol (icub-firmware and icub-firmware-shared). Todo: add it to the parser as well.

ale-git:
I need the help of @marcoaccame, because I added the 4 timeouts to the protocol but I was having a EMS crash at startup, and after many trials I reckoned that the maximum rop size was 384 bytes, while we need 392 now, both in EMS and yarprobotinterface as well.

https://github.com/ale-git/icub-firmware-shared/blob/4d0961e4c146cbadb991caa32f5883f68cfae500/eth/embobj/plus/comm-v2/icub/EoMotionControl.h#L776

The yarprobotinterface parser works, anyway.

marcoaccame:
> I need the help of @marcoaccame, because I added the 4 timeouts to the protocol but I was having a EMS crash at startup, and after many trials I reckoned that the maximum rop size was 384 bytes, while we need 392 now, both in EMS and yarprobotinterface as well.
> 
> https://github.com/ale-git/icub-firmware-shared/blob/4d0961e4c146cbadb991caa32f5883f68cfae500/eth/embobj/plus/comm-v2/icub/EoMotionControl.h#L776
> 
> The yarprobotinterface parser works, anyway.

Hi @ale-git to increase the max ROP size is surely possible even if not simple. Another simpler solution that does not increases required space is to add the timeout inside [eOmc_setpoint_t](https://github.com/ale-git/icub-firmware-shared/blob/4d0961e4c146cbadb991caa32f5883f68cfae500/eth/embobj/plus/comm-v2/icub/EoMotionControl.h#L688-L693)

valegagge:
> > I need the help of @marcoaccame, because I added the 4 timeouts to the protocol but I was having a EMS crash at startup, and after many trials I reckoned that the maximum rop size was 384 bytes, while we need 392 now, both in EMS and yarprobotinterface as well.
> > https://github.com/ale-git/icub-firmware-shared/blob/4d0961e4c146cbadb991caa32f5883f68cfae500/eth/embobj/plus/comm-v2/icub/EoMotionControl.h#L776
> > The yarprobotinterface parser works, anyway.
> 
> Hi @ale-git to increase the max ROP size is surely possible even if not simple. Another simpler solution that does not increase required space is to add the timeout inside [eOmc_setpoint_t](https://github.com/ale-git/icub-firmware-shared/blob/4d0961e4c146cbadb991caa32f5883f68cfae500/eth/embobj/plus/comm-v2/icub/EoMotionControl.h#L688-L693)

see here for more info:https://github.com/robotology/icub-firmware/issues/614#issuecomment-2980784399

randaz81:
I'm requesting a meeting F2F about this PR.
It is not compliant with our protocol and the control modes documentation which forbids any implicit internal change of control mode. I disagree with this.

emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
```diff
@@ -475,6 +527,60 @@ BOOL Joint_check_faults(Joint* o)
             }
         }
         
+        if (o->fault_state.bits.torque_ref_timeout && !o->fault_state_prec.bits.torque_ref_timeout)
```
valegagge, line 530: Hi @ale-git ,
this part of code is fine, but you can make it more readable and less prone to error if outside of the if section you define the descriptor and send the error at the end of `if `section. In each `if `you can leave the initialization of `par16`.

In any case, the code is fine.

emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
```diff
@@ -379,6 +379,17 @@ BOOL JointSet_do_check_faults(JointSet* o)
     
     BOOL fault = FALSE;
     o->external_fault = FALSE;
+
```
valegagge, line 382: Please remove the commented code.

valegagge (approved):
LGTM. I left two comments to improve the code's readability.

emBODY/eBcode/arch-arm/embobj/plus/mc/Joint.c
ale-git: I can't do that because each `if` can rise and error, i.e. more than one error type can be detected in a single cycle. All I can do is to move outside the declaration of `eOerrmanDescriptor_t`.

valegagge (changes requested):
Hi @ale-git ,
I'm sorry if I request other changes, but in the past day I dealed with the test of this new feature on AMC (thanks to @AntonioViscomi for his help!!).

We noticed that the feature doesn't work on AMC. See [here ](https://github.com/robotology/icub-firmware/issues/627#issuecomment-3143885923) for more details.

In addition, I noticed that in this PR are missing:
- the update of AMC version
- The diagnostic parameters do not match those in the icub-main repository code.
Infact, the error message we got is  wrong: 
image

the fw and the icub-main swap the use of param16 and param64.

ale-git:
I forgot to push the versions update (done). The .hex images are already updated.

ale-git:
@valegagge I tested the timeouts and they works on the AMC in the same way of the EMS as expected.

valegagge (approved):
Thanks @ale-git for your contribution!! In end, we get the timeout feature! 🚀

PR #598, opened 2025-05-19
AMC-FOC: Align codegen files structure to amcbldc

mfussi66:
This PR changes the folder structure of the generated code for the amcfoc to match the structure of the amcbldc project. In this way, the codegen artifacts are removed, and the helper python script copier.py can be use to copy the files from icub-firmware-models easily. 

The resulting amcfoc.mot project was built successfully after updating the Arm Keil project files.


marcoaccame (approved):
vey nice, thanks.

PR #599, opened 2025-05-21
Fix calibration issue observed in the neck joints

MSECode:
This PR solves a problem that we have observed lately on the neck joints that can be described as follows:

- all 3 neck joints calibrate correctly
- set `neck_pitch` joint in IDLE
- move the joint outside its hardware limits triggering HW error for overcoming hw boundaries
- stop the `yarprobotinterface`
- move the joint back inside the hw boundaries
- restart the `yarprobotinterface`

Doing the previous steps we  were continuing to see the joint in hw fault at the restart even if the encoder was reading a position inside the limit. The only way to recover the problem was restarting again the `yri` or power cycle the boards.

This was due to the fact of not having the deinitialization of the motion control inside the agents `stop()` method. 
Thus we added that. 
Then, in order to keep the state machine of the whole `agents manager` clean, we moved the MC calls inside the specific agent, such as the foc or advfoc agent

Update also versions for amc and amc2c

Issue related: https://github.com/icub-tech-iit/study-icub-headedge/issues/238

`icub-firmware-build` related PR: https://github.com/robotology/icub-firmware-build/pull/209

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
-            {2, 21},
-            {2025, Month::Mar, Day::twentyeight, 17, 17}
+            {2, 22},
+            {2025, Month::Mar, Day::twentyone, 13, 13}
         },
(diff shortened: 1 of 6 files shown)


valegagge (approved):
Thanks @MSECode !
I guess you already validated your changes on the robot.

If yes, for me these changes are ok!

MSECode:
Yes, sure. I've tested them on the neck joints for `ergoCub002` since that is the only new neck that have.

PR #611, opened 2025-06-09
Improving the `theEncoderReader` on dual core boards

SanLordKevin:
Improving the object `theEncoderReader`.
Now it includes two checks in its verify function:

1) It checks whether the encoder type sent to the board is supported by the board, meaning the firmware is present to handle that specific encoder.

Amo encoder is not supported by the `amc`, on `yarprobotinterface` appears an error that remands to a generic error of the encoder; there is not the case of unsupported encoder.
On the debugger also appears the error message.

Image

 ####

2) It performs one read from the encoder to verify that data is being received.

In case of encoder detached from the board a message is sent to `yarprobotinterface.`
In this case it remands to an `AEA_READING` error. In case of other encoder it gives a generic encoder error.
On the debugger also appears the error message.

Image

####
Tested on a `amc lego setup`.

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
+++ b/emBODY/eBcode/arch-arm/board/amc/application/v2/cfg/theApplication_config.h
@@ -38,8 +38,8 @@ namespace embot { namespace app { namespace eth {
         .property =
         {
             Process::eApplication,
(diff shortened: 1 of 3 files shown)


marcoaccame (approved):
Hi @SanLordKevin, it seems OK. Thanks

PR #612, opened 2025-06-11
Update tests for amc

MSECode:
With the PR we are:
- Updating the amc test project
- Adding the test for the Vaux check
- Adding the test for the ETH communication check
- Adding the test for checking the version of the fw of the test project
- Adding the test for checking the spi and the position of the encoders connected to the 3 spi channels

cc: @Nicogene

### DIFF
--- a/emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
+++ b/emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
@@ -11,8 +11,16 @@
 #include "embot_core.h"
 #include "embot_hw_motor.h"
 #include "embot_hw_button.h"
+#include "embot_hw_eth.h"
+#include "embot_hw_encoder.h"
 
 #include "embot_app_theLEDmanager.h"
+#include "embot_app_eth_theErrorManager.h"
+
+// Firmware versions - No need to be shared across multiple files - constexpress so that evaluated just at compile time
+static constexpr uint8_t _fMajorVersion = 1;
+static constexpr uint8_t _fMinorVersion = 0;
+static constexpr uint8_t _fPatchVersion = 0;
 
 bool TestRunnerAmc_CM7::testCanComm(uint8_t *data)
 { 
@@ -22,7 +30,28 @@ bool TestRunnerAmc_CM7::testCanComm(uint8_t *data)
     return true; 
 }
 
-bool TestRunnerAmc_CM7::testFwVersion(){ return true; }
+// TODO - Japo - June 2025: For now it is enough to just test if the eth cip links are up or down. 
+// In a second time we can think to add an UDP thread as in amc/examples/udpdeomobasic
+bool TestRunnerAmc_CM7::testEthComm(uint8_t *data)
+{
+    embot::core::print("Called method from testRunnerAmc_CM7");
+    embot::core::wait(300* embot::core::time1millisec);	
+    bool link1isup = embot::hw::eth::islinkup(embot::hw::PHY::one);    
+    bool link2isup = embot::hw::eth::islinkup(embot::hw::PHY::two);
+    std::string msg = std::string("ETH link 1 is ") + (link1isup ? "UP" : "DOWN") + " ETH link 2 is " + (link2isup ? "UP" : "DOWN");
+    embot::core::print(msg + " @ " + embot::core::TimeFormatter(embot::core::now()).to_string()); 
+    data [0] = (link1isup && link2isup) ?  0xAA : 0xBB;     
+    return true; 
+}
+
+bool TestRunnerAmc_CM7::testFwVersion(uint8_t *data)
+{ 
+    // saving fw versions to data in order
+    data[0] = _fMajorVersion;
+    data[1] = _fMinorVersion;
+    data[2] = _fPatchVersion;
+    return true; 
+}
 
 bool TestRunnerAmc_CM7::testLed(uint8_t on)
 {
@@ -99,7 +128,15 @@ bool TestRunnerAmc_CM7::testCin(uint8_t *data)
     return true; 
 }
 
-bool TestRunnerAmc_CM7::testVauxOK(){ return true; }
+bool TestRunnerAmc_CM7::testVauxOK(uint8_t *data)
+{ 
+    auto s = embot::hw::gpio::get(embot::hw::GPIO(embot::hw::GPIO::PORT::F, embot::hw::GPIO::PIN::five));
+    
+    if(s == embot::hw::gpio::State::SET){embot::core::print("OK"); data[0] = 0xAA;}
+    else{embot::core::print("NOK"); data[0] = 0xBB;}
+    
+    return true; 
+}
 
 bool TestRunnerAmc_CM7::testPwrnFailOK(){ return true; }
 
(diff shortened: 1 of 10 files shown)


emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -22,7 +30,28 @@ bool TestRunnerAmc_CM7::testCanComm(uint8_t *data)
     return true; 
 }
 
-bool TestRunnerAmc_CM7::testFwVersion(){ return true; }
+// TODO - Japo - June 2025: For now it is enough to just test if the eth cip links are up or down. 
+// In a second time we can think to add an UDP thread as in amc/examples/udpdeomobasic
+bool TestRunnerAmc_CM7::testEthComm(uint8_t *data)
+{
+    embot::core::print("Called method from testRunnerAmc_CM7");
```
valegagge, line 37: Do you still use this print for debug purposes?

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -22,7 +30,28 @@ bool TestRunnerAmc_CM7::testCanComm(uint8_t *data)
...
+    embot::core::wait(300* embot::core::time1millisec);
```
valegagge, line 38: Why do you need to wait 300 ms? Could you add a comment?

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -22,7 +30,28 @@ bool TestRunnerAmc_CM7::testCanComm(uint8_t *data)
...
+// In a second time we can think to add an UDP thread as in amc/examples/udpdeomobasic
+bool TestRunnerAmc_CM7::testEthComm(uint8_t *data)
+{
+    embot::core::print("Called method from testRunnerAmc_CM7");
+    embot::core::wait(300* embot::core::time1millisec);	
+    bool link1isup = embot::hw::eth::islinkup(embot::hw::PHY::one);    
+    bool link2isup = embot::hw::eth::islinkup(embot::hw::PHY::two);
+    std::string msg = std::string("ETH link 1 is ") + (link1isup ? "UP" : "DOWN") + " ETH link 2 is " + (link2isup ? "UP" : "DOWN");
+    embot::core::print(msg + " @ " + embot::core::TimeFormatter(embot::core::now()).to_string()); 
+    data [0] = (link1isup && link2isup) ?  0xAA : 0xBB;
```
valegagge, line 43: What do you think of defining the values `0xAA` and` 0xBB` in some point of your code as const with a name that makes their usage clear? If I'm not wrong, these values represent the return of the test and they have been defined in the protocol of the test suite.

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -163,6 +200,101 @@ bool TestRunnerAmc_CM7::testFault(uint8_t on, uint8_t *data)
...
+        
+    // result value for stopping flow if error before end of test
+    // same type as for methods in embot::hw::encoder
+    volatile embot::hw::result_t hwres {embot::hw::resNOK};
+        
+    // Initialize Encoders
+    for(size_t i=0; i<nENCs; i++)
+    {
+        embot::hw::ENCODER enc {static_cast<embot::hw::ENCODER>(i)};
+        embot::hw::encoder::init(enc, cfgAEA3);
```
valegagge, line 227: Why don't you use the return value of the init function to initialize done values?

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -163,6 +200,101 @@ bool TestRunnerAmc_CM7::testFault(uint8_t on, uint8_t *data)
...
+        encs[i] = enc;
+        //thecallbacks[i] = {onreadingdone, &encs[i]};
+        done[i] = true;
+    }
+    
+    // Start the data array with some initial values
+    // to be sure that if we are reading 0xFFFF we have actually problem with the chip
+    // and not with initialization of random values
+    for(size_t i=0; i<nENCs; i++)
```
valegagge, line 236: Here you should check the `done `vector before calling the start function, ensuring you call the start function only on the correctly initialized encoder.

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -163,6 +200,101 @@ bool TestRunnerAmc_CM7::testFault(uint8_t on, uint8_t *data)
...
+    {
+        spiRawValues[i] = 60666;
+        
+        // we want to be sure that our embot::hw::ENCODER objects have been intialized
+        embot::hw::ENCODER enc {static_cast<embot::hw::ENCODER>(i)};
+        // start non-blocking reading process (??stopped by callback??)
+        hwres = embot::hw::encoder::startRead(enc, {});
```
valegagge, line 243: In this way, hwres always contains the result of the last call (encoder num nENCS-1)

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -163,6 +200,101 @@ bool TestRunnerAmc_CM7::testFault(uint8_t on, uint8_t *data)
...
+    }
+    
+    if(hwres != embot::hw::resOK)
```
valegagge, line 246: Is it possible to improve the answer 0xBB, adding the value of the failed encoder?

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
```diff
@@ -163,6 +200,101 @@ bool TestRunnerAmc_CM7::testFault(uint8_t on, uint8_t *data)
...
+    {
+        embot::core::print("AEA3 reading NOK"); 
+        data[0] = 0xBB;
+        return false;
+    }
+    
+    // Get the spi encoder value
+    // Set a timeput for letting the reading process complete before getting the raw value
+    constexpr embot::core::Time timeout {200*embot::core::time1microsec};
```
valegagge, line 255: Why here you wait 200 ms instead of relying on the callback function?

valegagge (changes requested):
@MSECode Thanks for this! I have some questions and suggested changes, as certain implementation choices aren't entirely clear to me. Please take a look at the individual comments I've left. You'll notice some replicated situations, but I've only added a comment to one

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
MSECode: I just wait for a while in order to be sure that the ETH comm is actually responding.
I'll add a comment

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
MSECode: I did not understand. We can revise it later on.

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
MSECode: nope. StartRead returns a res_OK or res_NOK type of data

MSECode: I want to check if the encoder initialize the read correctly

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
MSECode: I was thinking to do that.
Check the issue: https://github.com/icub-tech-iit/ipts/issues/56#issuecomment-2962331466

valegagge (approved):
Thanks! LGTM

emBODY/eBcode/arch-arm/board/amc/test/src/testRunnerAmc_CM7.cpp
valegagge: Hi @MSECode, I guess you didn't plan this improvement in this PR, did you?

MSECode: No, I decided to add that in upcoming PRs, when we will have the first final version of the tests ready. For now I'm focused in making all test working and I prefer to keep the return value explicit in all test to be sure each test is returning a correct value without thinking to possible overriding of the values

PR #618, opened 2025-07-01
Kbemf added in current control

ale-git:
- added Kff parameter in CAN protocol for 2FOC velocity-current loop
- added Kbemf parameter in CAN protocol for 2FOC current loop 2FOC:
- added feed forward term Kff*vel_ref in velocity-current loop
- added feed forward term Kbemf*vel_meas in current loop
- velocity is measured at 20kHz on a 20 samples moving window


emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/include/2FOC.h
```diff
@@ -72,8 +72,8 @@ volatile extern int Ipeak;
 extern void DisableDrive(void);
 extern void I2Twatcher(void);
 
-extern void setIPid(int kp,int ki, char shift);
-extern void setSPid(int kp,int ki, char shift);
+extern void setIPid(int kp,int ki, int re, char shift);
```
isorrentino, line 75: Maybe here we can use the same name for the parameter `re` used in the .c file (`kbemf`)

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c
```diff
@@ -229,19 +231,27 @@ void setMaxTemperature(int peak)
...
+    IKbemf = kbemf;
+    IKs    = shift;
+    
+    if (ki == 0) ZeroControlReferences();
+                    
     IIntLimit = ((long)PWM_MAX)<<shift;
 }
 
-void setSPid(int kp, int ki, char shift)
+void setSPid(int kp, int ki, int kf, char shift)
```
isorrentino, line 246: Is the `kf` the `kff`? Maybe we can use `kff`

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c
```diff
@@ -1327,7 +1416,7 @@ int main(void)
         }
     }
 
-    setSPid(SKp, SKi, SKs);
+    setSPid(SKp, SKi, SKf, SKs);
```
isorrentino, line 1419: Is the `SKf` the `SKff`? Maybe we can use `SKff`

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/can_icubProto_parser.c
```diff
@@ -354,7 +384,11 @@ static int s_canIcubProtoParser_parse_pollingMsg(tCanData *rxpayload, unsigned c
         
                 if (max < power)
                 {
-                    setSPid((int)(fkp*32768.0f/power),(int)(fki*32768.0f/power),15-exponent);
+                    int kp = (int)(fkp*32768.0f/power);
+                    int ki = (int)(fki*32768.0f/power);
+                    int ke = (int)(fkf*32768.0f/power);
```
isorrentino, line 389: Is this the `kff` instead of the `ke`?

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/can_icubProto_parser.c
```diff
@@ -354,7 +384,11 @@ static int s_canIcubProtoParser_parse_pollingMsg(tCanData *rxpayload, unsigned c
...
+                    
+                    setSPid(kp,ki,ke,15-exponent);
```
isorrentino, line 391: Same here, is the `ke` the `kff`?

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/include/2FOC.h
```diff
@@ -72,8 +72,8 @@ volatile extern int Ipeak;
...
+extern void setSPid(int kp,int ki, int ke, char shift);
```
isorrentino, line 76: Here is the `ke` the `kff`? Maybe we could give the variable a descriptive name, one that already conveys what it holds, so we avoid any confusion later on.

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c
```diff
@@ -305,6 +315,67 @@ void ResetSetpointWatchdog()
...
+            
+            init_samples = FALSE;
+        }
+        
+        // gQEVelocity = (1 + gQEVelocity + gQEPosition - samples_circ_buffer[head]) / 2;        
+        gQEVelocity = gQEPosition - samples_circ_buffer[head];
+        samples_circ_buffer[head++] = gQEPosition;
+        head %= UNDERSAMPLING;
+        
+        if (++vel_loop_clock >= 20)
```
isorrentino, line 358: We could use `UNDERSAMPLING` instead of `20`.

emBODY/eBcode/arch-dspic/board/2foc/appl/2FOC-V3/src/2FOC.c
```diff
@@ -759,25 +841,25 @@ void __attribute__((__interrupt__, no_auto_psv)) _DMA0Interrupt(void)
...
                 if (gControlMode == icubCanProto_controlmode_speed_current)
                 {
-                    Is += (((long) (speed_error - speed_error_old)) << 4) + (long) (speed_error + speed_error_old);
-
-                    if (Is > Ipeak) Is = Ipeak; else if (Is < -Ipeak) Is = -Ipeak;
+                    // alternative formulation with ff term
+                    IsA += __builtin_mulss(speed_error+speed_error_old,SKi);
 
-                    if (starting)
+                    long IsF = __builtin_mulss(speed_error,SKp) + __builtin_mulss(SKf,CtrlReferences.WRef);
```
isorrentino, line 847: These modifications (from 835 to 862) are related to the velocity control loop. Do we want to keep them in this PR, or do we want to add them in a separate PR?
cc @valegagge

ale-git: I'm in favor of keeping it, since we also modified the velocity calculation.

isorrentino:
Some other minor comments.

valegagge:
## ⚠ Protocol compatibility 

Regarding the changes to the protocol, I checked the amc-bldc and amc-foc boards and they used the legacy  PID message format:

| AMC-BLDC                                                                                  | AMC-FOC                                                                                   |
|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| Image | Image |


If I understand correctly, you are suggesting using the PID.d parameter to store the kbemf value. So we don't need to change either the parser in icub-main and the EMS board, do we?

isorrentino:
Hi @valegagge we have repurposed the `Kd` field so it now carries `kbemf` for the current loop and `kff` for the velocity loop. This lets us keep the existing PID CAN message unchanged. I’m no low-level-protocol expert, but if the proposed solution meets the requirements, we can use it for now. If we ever need to transmit extra terms, such as a `kff` for the current controller, we’ll have to extend the CAN message protocol for the PID.

valegagge:
I put this PR in draft because we cannot merge the changes in the motor.c file since the AMc family boards rely on that. In addition, in that file, we changed the output of velocity from PWM to current, and we need to make it configurable.

isorrentino:
Hi @valegagge can we find a fast solution to have this PR merged instead of putting it in draft? 

Anyway I required an additional change to @ale-git as the modification to have the velocity-via-current enabled should not be part of this PR (https://github.com/robotology/icub-firmware/pull/618#discussion_r2185619570). It will be handled properly in a future PR as we already discussed during the dedicated meeting.

marcoaccame:
Hi @isorrentino, maybe we can find a solution. I will come back shortly.

marcoaccame:
> Hi @valegagge we have repurposed the `Kd` field so it now carries `kbemf` for the current loop and `kff` for the velocity loop. This lets us keep the existing PID CAN message unchanged. I’m no low-level-protocol expert, but if the proposed solution meets the requirements, we can use it for now. If we ever need to transmit extra terms, such as a `kff` for the current controller, we’ll have to extend the CAN message protocol for the PID.

It is possible to keep the CAN protocol as it is now and use the **SET_VELOCITY_PID.KD** for `Kff` and **SET_CURRENT_PID.KD** for `Kbemf`. I have verified w/ @mfussi66 that the MBD motor control boards (`amcbldc`, second core of `amc `and `amcfoc`) don't use the **KD** field. For sure, when these MBD boards need these values there are changes to be done.

My design preference would be to go directly to have new dedicated CAN message(s) and not to cannibalize existing ones, but I understand that as a matter of urgency we can go through this intermediate step.

cc @pattacini @maggia80 

In any case, there are some review comments that I have prepared for this PR for the case of keeping existing CAN messages as a temporary solution. 

I will post it shortly.

marcoaccame (commented):
### Let me summarize what I understand

From the bottom up to `yarprobotinterface`and xml files.

We want the `2foc` board to manage two extra parameters (the `Kbemf` and the `Kff`) that so far are not in the CAN protocol.

The idea is to fill them inside two existing CAN messages, so that:

- the **KD** field of  **SET_CURRENT_PID** contains `Kbemf`,
- the **KD** field of **SET_VELOCITY_PID** contains `Kff`.

See in [here](https://github.com/robotology/icub-firmware/blob/master/emBODY/eBdocs/arch-arm/TSD-ICUBUNIT-canprotocol-motorboards.docx) the CAN protocol format for these two messages (sections 2.3.5 and 2.3.7)

And we have changes in the `2foc` code to extract the above fields and use them. It is OK to do that for the 2foc because it does not use the derivative term for current and velocity PIDs. 

Going up on the stack, in `Motor.c`, used by all ETH boards (`ems`, `amc`, `amcfoc`), there are changes that fill the **KD** fields of **SET_CURRENT_PID** and  **SET_VELOCITY_PID** accordingly. In particular:

- the `Kff` param is taken from `eOmc_motor_config_t::pidvelocity::kff` and placed into **SET_VELOCITY_PID.KD**,
- the `Kbemf` param is taken from `eOmc_motor_config_t::pidcurrent::kff` and placed into **SET_CURRENT_PID.KD**.

So, the `eOmc_PID_t::kff` field is interpreted either as feed forward gain (correctly) or as back EMF. 

This `eOmc_motor_config_t` is prepared and transmitted by `yarprobotinteface` that reads from xml files values such as [these](https://github.com/robotology/robots-configuration/blob/master/ergoCubSN002/hardware/motorControl/left_leg-eb9-j4_5-mc.xml#L97-L121) whose template is in [here](https://github.com/robotology/robots-configuration/blob/master/iCubTemplates/iCubTemplateV6_0/hardware/motorControl/body_part--ebX-jA_B-mc.xml#L149-L174). The template already contains `kff` and tells that `kd` is a derivative gain.



### And my comment

The motor control boards that used MBD (`amcbldc`, second core of `amc` and `amcfoc`) do not use the  **SET_VELOCITY_PID.KD** and  **SET_CURRENT_PID.KD** fields, so this use of the CAN protocol does not do any harm. See also comment of @valegagge. cc @pattacini @mfussi66

So, also `Motor.c` can be OK as long as we don't modify `Motor_set_run()`  w/ `icubCanProto_controlmode_speed_current`. Correct @valegagge ?

I don't comment on the internals of the 2foc code that has been already revised.

So far so good, but I recommend some changes for clarity and to avoid confusion.



### Recommended changes

1. Add a `kbmf` in `eOmc_PID_t` and in the xml file documentation and add its parsing in `yarprobotinterface`. It is not a big effort, avoids confusion and it surely must be done rather soon.

2. As an alternative of point 1: write in `Motor.c` a comment that the use of `pidcurrent->kff` to initialize `kbemf` is a  temporary workaround due to the lack of a variable that soon will be added. 
3. As an alternative to point 1: write in the template files of robots-configuration that the `kd` term is not used by 2foc and other motor control boards and contains a feed forward or a back EMF.
4. Write a comment in `Motor_config_current_PID_2FOC()` and in `Motor_config_velocity_PID_2FOC()` that the `Kd` is nit used by the supported motor boards and that it is safe to use the field for another purpose.

5. remove the `icubCanProto_controlmode_speed_current` in `Motor_set_run()`.

6. produce binaries for at least the `ems` and `amc` (`mc4plus` and `mc2plus` don't use the changed code)

My preference is for points 1, 4, 5, 6

### After that

Surely, after that I would think of dedicated new CAN message for feed forward, back EMF and whatever is needed.

pattacini:
Sorry, I was sort of off these days for an important project demo.

>I have verified w/ @mfussi66 that the MBD motor control boards (amcbldc, second core of amc and amcfoc) don't use the KD field.

Uhm, my understanding is the opposite. This modification would be disruptive for the new electronics.
I've quickly verified the architectural model as @valegagge did [above](https://github.com/robotology/icub-firmware/pull/618?notification_referrer_id=NT_kwDOADkJ1rMxNzMyNzk4ODQ4MTozNzM4MDcw#issuecomment-3032311647) and we do use D for both velocity and current PID's. They're initialized from the CAN network.

Did you mean that the set velocity/current D is not executed upstream by AMC/AMCFOC?
If so, why?

My understanding is so different than yours, @marcoaccame @mfussi66, that I might be missing something important here and thus we definitely need to wrap it up.

At any rate, we should certainly address two contrasting requirements:
- Avoid impacting the new electronics with this change[^1].
- Avoid curbing the fast-paced developments on 2FOC.

The ideal solution would entail expanding on the CAN protocol. This would surely take some time, although we will have to get it done, eventually.

We're to find intermediate and viable solutions.

Given that fast-paced devs will probably hit again on the overall codebase that has several shared parts, I think it's worth considering branching out the codebase of the new electronics (i.e., AMC, AMCBLDC, AMCFOC) from the codebase of the current electronics (i.e., EMS, 2FOC, MC4Plus).
The only thing they share is actually some parts of the codebase since they fundamentally differ for roadmaps, approaches, electronics, performance, development paces (external vs. internal).

This may sound a daunting task now, but I could honestly see benefits that can rapidly pay off.
Of course, the task could be completed in steps, being the first aimed at untangling this issue.

[^1]: I'm following the hypothesis I made but I'd be more than happy to be proven wrong. Also, if this hypothesis holds, we could think of commenting out those lines in the FSM handling the D term in the architectural model of the new electronics and assume D to be always null (temporarily).

maggia80:
> Given that fast-paced devs will probably hit again on the overall codebase that has several shared parts, I think it's worth considering branching out

yes!!!! Let's do it!!!

marcoaccame (approved):
Hi @ale-git the PR is fine and if you think that the tests you mention in [here](https://github.com/robotology/icub-firmware/issues/637) are OK pls move the PR in ready for review and we can start to merge w/ this order:
- this PR;
- a PR yet to be prepared in https://github.com/robotology/icub-firmware-build for the binaries of 2foc, ems, mc4plus, mc2plus, amc;
- https://github.com/robotology/icub-firmware-shared/pull/119;
- https://github.com/robotology/icub-main/pull/1040;
- https://github.com/robotology/robots-configuration/pull/752.

ale-git:
Bugfix committed.

valegagge:
Hi @ale-git, are you sure to delete all the CAN protocol files in the 2foc fw?

ale-git:
> Hi @ale-git, are you sure to delete all the CAN protocol files in the 2foc fw?
> 
> Yes, they were a local copy. They must be included from their proper location instead.

valegagge (approved):
LGTM! thanks

PR #632, opened 2025-08-18
amcfoc temperature readings implementation

SanLordKevin:
`amcfoc` temperature readings implementation:
- core temperature 
- amcfoc-Tdriver temperatures

Bugfixed the use of the ADC buffer in the callback functions `Adc3DmaComplete_callback`.
Ordered the code.


marcoaccame (approved):
ok, thanks

PR #633, opened 2025-08-22
Finalize tests for amc and amc2c

MSECode:
This PR adds the bsp for the i2c driver in the amc and finalize both amc and amc2c test fw for the ipts suite v1.6.1

Add i2c driver to amc bsp
Update test fw for i2c and bsp related
Added dma in i2c
Update amc cm7 test fw version and add version to cm4 test fw
Update vin threshold


marcoaccame (approved):
ok, thanks

PR #639, opened 2025-09-08
Increase wait on initial tests for amc cm7 tests

MSECode:
This PR adds the following fixes:
- increases the waiting for test on amc cm7 core 
- adjust the PWM thresholds on the cm4 test procedure since the amc fed correctly the requested percentage of PWM (differently from amc_bldc. Thus it is better to check `amc_bldc` bsp)
- small adjustment on cin threshold as well


valegagge (approved):
LGTM

PR #649, opened 2025-10-07
Add Calibration, Trajectory Planner in the AMCFOC

RaffaelePumpo:
- File MBD updated with latest version of icub-firmware-models. 
- Added Rotor-Stator Calibration
- Added Trajectory Planner for Control Mode Position


marcoaccame (approved):
Jolly good. Thanks @RaffaelePumpo

PR #668, opened 2026-02-25
EMS ver 3.112: change severity aksim crc error to info

valegagge:
In this patch, we change the severity of CRC errors for the Aksim encoder managed by EMS. The severity is now set to INFO, preventing unnecessary alarms for the user.


valegagge:
cc @Nicogene

PR #679, opened 2026-05-22
UDP test project for amcmj1 and amcfoc

SanLordKevin:
Added a project for the core CM4 used in master mode that initialize a stack UDP. Tested by pinging the board and by using `netcat` to start a UDP connection.

Note: at the moment in these project `embot::core::now()` is broken, so to run the project on the board you have to disable the `EEPROM`, as its `init` function calls it.


marcoaccame:
> Note: at the moment in these project `embot::core::now()` is broken, so to run the project on the board you have to disable the `EEPROM`, as its `init` function calls it.

Ok, thanks @SanLordKevin for this information. The development on `amcmj1` is now WIP and I will soon have a look at how `embot::core::now()` behaves.

marcoaccame (approved):
@SanLordKevin very good that we have UDP running on the `amcmj1`.

PR #681, opened 2026-06-03
amcmj1 updater and maintainer

SanLordKevin:
Adding updater and maintainer for amcmj1. Both project can run using cm7 or cm4 as master.


marcoaccame (approved):
ok, thank

PR #682, opened 2026-06-03
Updates for fix of calibration type 7 absEncoder-EoEncoderReader sign inconsistency

MSECode:
This PR aims at fixing a problem, more a corner case problem, that might generate during the calibration of the hand joints that use calibration `type7` and `type6`. The problem can be detailed shortly as follows:

**Problem / Root Cause**

Calibration type 7 could underflow in the Hall ADC reader path because the reader removes the sign from the conversion factor, applies an unsigned offset subtraction, and only later the abs encoder reintroduces the sign. Near endpoint values, this could wrap the normalized encoder reading.

Calibration type 6 had a similar edge risk with MAIS readings: the normalized MAIS value could exceed the `uint16_t` range before reaching `AbsEncoder_update()`, depending on raw MAIS limits and current encoder position.

**Solution**

For type 7, the calibration now computes the reader offset using the absolute resolution and a safe margin, selecting the correct endpoint depending on the resolution sign. This keeps the normalized Hall ADC value away from unsigned underflow/overflow boundaries.

For type 6, MAIS scaling now supports an offset, and calibration applies the same safe-margin approach before values reach the abs encoder path. The MAIS reader also clamps normalized values to the valid `uint16_t` range to prevent wrapping.

The following diagram can help in the points where the problem generated before the update:

```mermaid
flowchart TD
    A["JointSet_calibrate()\n\ncomputedResolution = (vmax - vmin) / (jmax - jmin)\nabsResolution = abs(computedResolution)\noffset = endpoint / absResolution - safeMargin\ncomputedZero = sign-dependent zero"]

    B["embot::app::eth::theEncoderReader::Scale()\n\nScale(target, scaler)\nscaler.scale = computedResolution\nscaler.offset = offset"]

    C["embot_app_eth_theEncoderReaderLegacy.cpp\nImpl::Scale()\n\neOencoderreader_Scaler sca {scaler.scale, scaler.offset}\neo_encoderreader_Scale(..., &sca)"]

    D["EOtheEncoderReader.c\neo_encoderreader_Scale()\n\nif encoder is absanalog:\n  UpdatedHallAdcConversionFactors(scale)\n  UpdatedHallAdcOffset(offset)\n\nif encoder is MAIS:\n  UpdatedMaisConversionFactors(scale)"]

    E["EOappEncodersReader.c\neo_appEncReader_UpdatedHallAdcConversionFactors()\n\nhallAdcConversionData.factors[jomo] = convFactor"]

    F["EOappEncodersReader.c\neo_appEncReader_UpdatedHallAdcOffset()\n\nhallAdcConversionData.offsets[jomo] = offset"]

    G["EOappEncodersReader.c\ns_eo_appEncReader_hallAdc_rescale2icubdegrees()\n\nuint32_t retval = val_raw\nfloat divider = factor\nif(divider < 0) divider = -divider\nretval = (float)retval / divider\nretval -= offset"]

    H["EOtheMotionController.c\ns_eo_motioncontrol_updatePositionFromEncoder()\n\nMController_update_absEncoder_fbk(index, valueinfo->value)"]

    I["Controller.c\nMController_update_absEncoder_fbk()\n\nAbsEncoder_update(enc + k, (uint16_t)positions[k])"]

    J["AbsEncoder.c\nAbsEncoder_update(uint16_t position)\n\nposition_sure = position\nint16_t delta = position - position_sure\ndistance += delta"]

    K["AbsEncoder.c\nAbsEncoder_position()\n\nposition = mul * distance / div - zero"]

    A -->|"float scale\nint32_t offset"| B
    B -->|"Scaler {float scale, int32_t offset}"| C
    C -->|"eOencoderreader_Scaler"| D
    D -->|"absanalog scale"| E
    D -->|"absanalog offset"| F
    E -->|"float factor"| G
    F -->|"int32_t offset"| G
    G -->|"uint32_t value[0]\nunderflow can happen here"| H
    H -->|"uint32_t* positions"| I
    I -->|"cast to uint16_t\nwrap can happen here"| J
    J -->|"int32_t distance"| K
```

This PR is related to:
- icub-firmware-shared: https://github.com/robotology/icub-firmware-shared/pull/125
- icub-main: https://github.com/robotology/icub-main/pull/1065
- icub-firmware-build: https://github.com/robotology/icub-firmware-build/pull/234

### DIFF
--- a/emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
+++ b/emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
@@ -2018,16 +2018,43 @@ void JointSet_calibrate(JointSet* o, uint8_t e, eOmc_calibrator_t *calibrator)
             o->joint[e].running_calibration.data.type7.state = calibtype7_st_inited;
             o->joint[e].running_calibration.data.type7.is_active = TRUE;
             
-//            o->joint[e].calib_type7_data.is_active = TRUE;
-//            o->joint[e].calib_type7_data.state = calibtype7_st_inited;
-//            o->joint[e].calibration_in_progress = (eOmc_calibration_type_t)calibrator->type;
-            
-            
             // 2) calculate new joint encoder factor and param_zero
             eOmc_joint_config_t *jconfig = &o->joint[e].eo_joint_ptr->config;
             float computedJntEncoderResolution = (float)(calibrator->params.type7.vmax - calibrator->params.type7.vmin) / (float) (jconfig->userlimits.max  - jconfig->userlimits.min);
-            
-            int32_t offset = (((float)calibrator->params.type7.vmin)/computedJntEncoderResolution) - jconfig->userlimits.min;
+            float absComputedJntEncoderResolution = (computedJntEncoderResolution >= 0) ? computedJntEncoderResolution : -computedJntEncoderResolution;
+
+            float jntRange = (float)(jconfig->userlimits.max - jconfig->userlimits.min);
+            // Keep the normalized reader value away from zero to avoid unsigned underflow on endpoint drift.
+            // Same margin used by the type 6/7 calibration limit check: 80 degrees in iCubDegrees.
+            const int32_t normalizedDistanceMargin = 14563;
+            int32_t offset = 0;
+            int32_t computedJntEncoderZero = 0;
+
+            if((0.0f == absComputedJntEncoderResolution) || (jntRange <= 0.0f) || ((jntRange + 2.0f*normalizedDistanceMargin) >= 65535.0f))
+            {
+                char info[70];
+                snprintf(info, 70, "calib7: invalid params");
+                JointSet_send_debug_message(info, e, 0, 0);
+                return;
+            }
+
+            if(computedJntEncoderResolution >= 0)
+            {
+                offset = (((float)calibrator->params.type7.vmin) / absComputedJntEncoderResolution) - normalizedDistanceMargin;
+                computedJntEncoderZero =  normalizedDistanceMargin -((int32_t)jconfig->userlimits.min);
+            }
+            else
+            {
+                offset = (((float)calibrator->params.type7.vmax) / absComputedJntEncoderResolution) - normalizedDistanceMargin;
+                computedJntEncoderZero = -((int32_t)jconfig->userlimits.max) - normalizedDistanceMargin;
+            }
+
+            {
+                char info[80];
+                snprintf(info, sizeof(info), "calib7:T res%.4f off%d z%d m%d", computedJntEncoderResolution, offset, computedJntEncoderZero, normalizedDistanceMargin);
+                JointSet_send_debug_message(info, e, 0, 0);
+            }
+           
             
 
 #if 1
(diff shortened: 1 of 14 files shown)


MSECode:
May I add to the PR the removal of the lines between the `else` of this MACRO since they are never called: https://github.com/MSECode/icub-firmware/blob/33402fe6e29952745d8927bab754d68dc98734ad/emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c#L2097-L2122

marcoaccame:
Hi @MSECode, I see that the PRs in icub-main and icub-firmware-shared are to be changed to merge into master and that this one and the icub-firmware-build are vs devel and still in draft. Would it be better they also are vs master? In any case, drop me a message when everything is ready.

valegagge:
I started the review, but I couldn't complete it here now. Therefore, I just dropped a reference to the old code where the round problem was taken into account.

https://github.com/robotology/icub-firmware/commit/4691befe36d8ffdcf58b7177884a018cf91e5bc5

Asap, I'll go ahead with the review.

emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
```diff
@@ -1963,28 +1963,56 @@ void JointSet_calibrate(JointSet* o, uint8_t e, eOmc_calibrator_t *calibrator)
...
+                
+                // 4) calculate a sage margin to avoid unsigned overflow/underflow on endpoint drift.
+                // Keep the normalized reader value away from zero to avoid unsigned underflow on endpoint drift.
+                // Same margin used by the type 6/7 calibration limit check: 80 degrees in iCubDegrees.
+                float jntRange = (float)(jconfig->userlimits.max - jconfig->userlimits.min);
+                const float maxMargin = (65535.0f - jntRange) / 2.0f;
+                const float normalizedDistanceSafety = 182.0f;
+                const int32_t maxNormalizedDistanceMargin = (int32_t)(maxMargin - normalizedDistanceSafety);
+                
+                int32_t normalizedDistanceMargin = CALIB_TYPE_6_7_POS_ERROR_TRHESHOLD;
```
valegagge, line 1976: Hi @MSECode , 
why here you use `CALIB_TYPE_6_7_POS_ERROR_TRHESHOLD ` (~= 80 deg), instead of the `65532 `(~=360 deg) divided by two?

valegagge (approved):
Hi @MSECode, good work! 
Please see a comment in the code directly.

The last question: during your work and testing, did you have the chance to check if the use of `CALIB_TYPE_6_7_POS_ERROR_TRHESHOLD ` in `JointSet_calibType6_check_reached_pos` is correct, or if its value is too high?
thanks

emBODY/eBcode/arch-arm/embobj/plus/mc/JointSet.c
MSECode: I've used that since it is the same margin used for checking the exceeding of the hw boundaries, thus just to be compliant with that.
Now, the reason why we have that value as margin for the hw boundaries is not totally clear to me.
I just used it since it dates to quite lots of time ago and I supposed there was an exact reason to it, which I actually did not investigate much.
The macro was initially defined in here: https://github.com/robotology/icub-firmware/pull/682/changes#diff-4b2e4ea9299182280d34d8772d5ea78a539822f6db3b2c84914ef66601dee210L59
I've moved to the header file just to use it in another Class

valegagge: Yes, I saw you just moved the definition of three macros. Its fine!
You are right. The code you are refering to is old, but I guess I used so a big range to let the joint calibrate. 
I have concerns that the margin is too high and could cause issues during calibration. 
In any case, we can proceed as is, and during the test we can keep an eye on this unusual threshold. 

Does it work for you?

MSECode: Totally agree

valegagge:
Hi @MSECode ,
As @marcoaccame [wrote](https://github.com/robotology/icub-firmware/pull/682#issuecomment-4631799364), do you think thart this fix should be merged vs `Master` or `Devel`?

MSECode:
Yes, that's correct. We might merge directly to master. On top of that the latest updates from devel should not be related to our changes.
I will rebuilt all of the .hex to be sure we have correctly the fw aligned with master.
Since then, all the fw contains fixes it does make more sense to merge it to master.

PR #687, opened 2026-06-17
adding amcmj1 yri application 

SanLordKevin:
adding amcmj1 application that talks with yarprobotinterface. This can run bpth n CM7 and CM4.
I tested this application with a force-torque sensor. 
I successfully tested the FT service (my FT is broken as you can see):

yarprobotinterface running happpily
Image

Value read from the port
Image


marcoaccame (approved):
ok, thanks @SanLordKevin

PR #690, opened 2026-06-19
Fix/amcx runningmode

SanLordKevin:
Adding for the amcmj1 and amcfoc the capability to handle the runningmode besteffort.
I tested on the amcmj1 application using the FT service.


marcoaccame:
ok, thanks.

marcoaccame (approved):
ok, thanks

PR #692, opened 2026-07-08
amcmj1: refactory config.h files for different projects

SanLordKevin:
Refactor  `_config.h` files for amcmj1 BSP for different projects. 
Every process has now its own BSP configuration files with their specific define.


marcoaccame (approved):
ok, @SanLordKevin thanks

PR #693, opened 2026-07-13
amcmj1 - improved FDCAN 

SanLordKevin:
Improved the `embot_hw_can.cpp` file to handle a change in `stm32h7xx-hal-driver`  to file  [stm32h7xx_hal_fdcan.h](https://github.com/STMicroelectronics/stm32h7xx-hal-driver/compare/v1.10.0...v1.11.5#diff-c2ba6df8cde8e9d30e9aa9281f30bc3015592a1a011e8e9283bb4a45fc379481) .

Improved the amcmj1 CAN BSP: configured the pin that enables the 5V on the CAN BUS, and the pin that reads the status of the chip that power the line. We do not use this last reading atm.
Added the `EMBOT_ENABLE_hw_can_5V` macro to enable the 5V independently from the `init` of the CAN peripheral. 

Improved the `xx_config.h` files, fixed scatter file for cm4 of app.mot., removed deinit from motor.


marcoaccame (approved):
OK, thanks

PR #696, opened 2026-07-31
amcmj1 - adding button and macro for disabling motor interrupt

SanLordKevin:
Starting from amc.app.mot I took the logic for the fault button and the handling of the IRQ linked to the fault.

Tested by: seeing the led associated to the fault becoming red without fault inserted and with fault pressed, led is off otherwise. 
Also using the debugger i could see this behaviour. I haven't noticed strange behaviour, but i ask @marcoaccame to also test this before merging.


marcoaccame (approved):
ok, thanks

PR #702, opened 2026-09-14
amcmj1: I2C BPS and ping test

SanLordKevin:
Adding amcmj1 I2C BSP and a ping test for the IMU onboard.
Worked on both cores using the test hw project.


marcoaccame (approved):
ok, thanks

PR #704, opened 2026-09-18
amcmj1: porting cm7 project to use lwip

SanLordKevin:
as per title. 
Tested loader, updater, maintaner and yri application.


marcoaccame (approved):
oki

